PR #4, opened 2019-02-27
Add basic support for the new license expression field

coffeewiththegrinds:
Adding support for some basic parsing of the new license expression field, which is found in more recent nuget packages.  [https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg#project-properties](url)


stevespringett:
Thanks for the PR

PR #8, opened 2019-07-15
Include license URL if license id is not available. Moved to 1.1 schema

scorgatelli-docutech:
Include the license URL when the license id is not available in the package metadata. This allows us to to further post-processing on the BOM to determine what the license for the package is. This is helpful since most nuget packages don't currently include the license id.

I had to bump the schema version to 1.1 to support this change.


stevespringett:
Thanks for the PR.

PR #9, opened 2019-07-23
Added option to recursively scan PackageReferences of included projects (.csproj support only).

MangoFloat:
- If you supply a single .csproj, it will scan the project reference tree and recursively retrieve all the package references of the included projects.
- All project references will be sent for scanning, producing a single BOM file for all recursive package references.


stevespringett:
Thanks for the PR.

stevespringett:
The travisci build passed, however, locally, I receive the following error when building:


> Program.cs(34,34): Warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. (CS1998) (CycloneDX)

MangoFloat:
> The travisci build passed, however, locally, I receive the following error when building:
> 
> > Program.cs(34,34): Warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. (CS1998) (CycloneDX)

It's a warning about one the methods that was added; I wouldn't worry about it, as it shouldn't affect anything, unless it stops you from building?

stevespringett:
I can execute build.sh just fine (no errors), although I do get the above warning. When in VS, doing a build (or rebuild) results in the above not allowing the build to continue. Perhaps its a VS config thing I can set (not really a .net developer so I'm not sure).

MangoFloat:
You might have treat warnings as errors enabled.
You can change the settings in VS if you right click the project -> Build -> Treat warnings as errors.

Issue #30, opened 2019-11-12
Licenses are not downloaded


SS-TT:
I've opened an issue on Dependency-Track and I've been asked to move it here.
Here's the link: DependencyTrack/dependency-track#510

stevespringett:
As noted in https://docs.microsoft.com/en-us/nuget/reference/nuspec#licenseurl, licenseUrl is deprecated. However, it continues to be widely used. 

The CycloneDX .NET implementation does not download the license file or attempt to identify it. It might be interesting to use the GitHub API to determine the open source license. They already have logic to identify ab out a dozen licenses (and small variances such as copyrights in MIT).

coderpatros:
It would be good to do this for opensource.org licence URLs too. It looks like, most of the time, the SPDX ID is in the URL. Maybe a way to implement this would be to extract the ID from the URL and cross reference it with the list of actual SPDX IDs?

coderpatros:
We have 600 references to the Apache licence URLs in our BoMs. So even a simple match on URL could be useful.

coderpatros:
Can this issue be closed now @SS-TT ?

SS-TT:
Sure, sorry for missing that.

PR #39, opened 2019-11-18
Added license handling using Github API when possible.

ST-Apps:
Potential fix for #30.
I don't see a contribution policy so I tried to keep things as close as possible to other classes in the project.

Waiting for feedback on this :)

### DIFF
--- a/CycloneDX/Services/GithubService.cs
+++ b/CycloneDX/Services/GithubService.cs
@@ -1,0 +1,88 @@
+﻿// This file is part of the CycloneDX Tool for .NET
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Copyright (c) Steve Springett. All Rights Reserved.
+
+using CycloneDX.Models;
+using Newtonsoft.Json;
+using System;
+using System.Net.Http;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using License = CycloneDX.Models.License;
+
+namespace CycloneDX.Services
+{
+
+    public interface IGithubService
+    {
+        Task<License> GetLicenseAsync(string licenseUrl);
+    }
+
+    class GithubService : IGithubService
+    {
+
+        private string _baseUrl;
+        private HttpClient _httpClient;
+        private Regex _repositoryIdRegex = new Regex(@"https?\:\/\/[^\/]+\/(?<repositoryId>[^\/]+\/[^\/]+)\/.+$");
+
+        public GithubService(HttpClient httpClient, string baseUrl = null)
+        {
+            _httpClient = httpClient;
+            _baseUrl = baseUrl ?? "https://api.github.com/";
+        }
+
+        public async Task<License> GetLicenseAsync(string licenseUrl)
+        {
+            // Regex may let other URLs pass, a plain filter on URLs containing "github" may be enough
+            if (licenseUrl == null || !licenseUrl.Contains("github")) return null;
+
+            // Detect correct repository id starting from URL
+            var match = _repositoryIdRegex.Match(licenseUrl);
+
+            // License is not on GitHub, we need to abort
+            if (!match.Success) return null;
+            var repositoryId = match.Groups["repositoryId"];
+
+            Console.WriteLine($"Retrieving GitHub license for repositoy {repositoryId}");
+
+            // Generate request URL and message
+            var githubLicenseUrl = $"{_baseUrl}repos/{repositoryId}/license";
+            var githubLicenseRequestMessage = new HttpRequestMessage(HttpMethod.Get, githubLicenseUrl);
+            // Add needed headers
+            githubLicenseRequestMessage.Headers.UserAgent.ParseAdd("CycloneDX/1.0");
+            githubLicenseRequestMessage.Headers.Accept.ParseAdd("application/json");
+            // Send HTTP request
+            var githubResponse = await _httpClient.SendAsync(githubLicenseRequestMessage);
+
+            if (githubResponse.IsSuccessStatusCode)
+            {
+                // License found, extract data
+                var githubLicense = JsonConvert.DeserializeObject<GithubLicenseRoot>(await githubResponse.Content.ReadAsStringAsync());
+                return new License
+                {
+                    Id = githubLicense.License.SpdxId,
+                    Name = githubLicense.License.Name,
+                    Url = licenseUrl
+                };
+            }
+            else
+            {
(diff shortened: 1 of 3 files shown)


stevespringett:
Thanks for the PR. I think the GitHub approach is valid, especially if the nuget package is using the legacy license_url instead of the newer license which supports SPDX expressions. 

Thoughts/concerns @coderpatros ?

CycloneDX/Services/GithubService.cs
```diff
@@ -0,0 +1,88 @@
...
+            _baseUrl = baseUrl ?? "https://api.github.com/";
+        }
+
+        public async Task<License> GetLicenseAsync(string licenseUrl)
+        {
+            // Regex may let other URLs pass, a plain filter on URLs containing "github" may be enough
+            if (licenseUrl == null || !licenseUrl.Contains("github")) return null;
+
+            // Detect correct repository id starting from URL
+            var match = _repositoryIdRegex.Match(licenseUrl);
```
coderpatros: @ST-Apps the regex to extract the repository id looks pretty relaxed. Is there a reason you haven't added the github check to the regex instead of the contains check?

CycloneDX/Services/GithubService.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            // License is not on GitHub, we need to abort
+            if (!match.Success) return null;
+            var repositoryId = match.Groups["repositoryId"];
+
+            Console.WriteLine($"Retrieving GitHub license for repositoy {repositoryId}");
+
+            // Generate request URL and message
+            var githubLicenseUrl = $"{_baseUrl}repos/{repositoryId}/license";
```
coderpatros: I have some concerns that this won't always return the correct licence. It looks to me that this will return the _current_ licence, not necessarily the _actual_ licence for previous versions of a package.

coderpatros:
@stevespringett besides the comments above it looks pretty good. Although some accompanying tests would be nice :)

CycloneDX/Services/GithubService.cs
ST-Apps: To be fair I was half asleep when I wrote it, sorry for that :)
I'm at work now and can't edit the code, but this one may be good enough to remove the `Contains`:

```
https?\:\/\/.*github.*[^\/]+\/(?<repositoryId>[^\/]+\/[^\/]+)\/.+$
```

I just did a quick test with [regex101](https://regex101.com/tests):

- https://github.com/CycloneDX/cyclonedx-dotnet/blob/master/LICENSE **matches** 
- https://raw.githubusercontent.com/CycloneDX/cyclonedx-dotnet/master/LICENSE **matches** 
- https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm **doesn't match** 

If you think it's fine, I'll push again when I'll be back at home.

CycloneDX/Services/GithubService.cs
ST-Apps: You're correct, but GitHub's APIs work only with `master` branch, meaning that you can only get the latest version of the `LICENSE` file.

This is the reference that I used: https://developer.github.com/v3/licenses/

To be fair, even if GitHub APIs allowed to get licenses from other branches, I don't think it would be enough to solve this issue. Let's say for example that NuGet returns `2.3.0` as version for a given package, how would you match it to a branch on the target repository to get its license? 

Also, let's see another example: NLog, on NuGet, has the very same license file linked for all its releases. This means that they're linking to `master/LICENSE.txt` for all of their versions, but what if the next version changes license? Well, all the older versions are still linked to the file on master, so it will look like they always had that license.

CycloneDX/Services/GithubService.cs
coderpatros: Yeah, know what it's like to be coding late and wanting to get something finished off :)

I was thinking of being really specific. I think something like this would cover >90% of cases...

```
^https?\:\/\/((github\.com)|(raw\.githubusercontent\.com))\/(?<repositoryId>[^\/]+\/[^\/]+)\/blob\/master\/LICENSE$
```

This would also address the below issue as we would only ever be resolving the current licence if it was referenced specifically.

CycloneDX/Services/GithubService.cs
coderpatros: That endpoint also has an optional `ref` parameter, i.e. [https://api.github.com/repos/coderpatros/dotnet-authenticated-httpclient/license?ref=v1.3.0](https://api.github.com/repos/coderpatros/dotnet-authenticated-httpclient/license?ref=v1.3.0)

Although I'd be surprised if there were many package maintainers who actually went to the trouble of correctly referencing their licence file per release. I know I don't :)

This is a corner case and I'd prefer to get this merged than handle all possible scenarios. But I think if we can't be sure it should fallback to current behaviour to prevent reporting incorrect licence information.

CycloneDX/Services/GithubService.cs
ST-Apps: Updated the regex starting from yours.
`blob` is not present on `raw.githubusercontent.com`, and `LICENSE` file can also have an extension so I allowed characters after `LICENSE`.

CycloneDX/Services/GithubService.cs
ST-Apps: That `ref` parameter is the key, I don't know how I missed it.
I just pushed an update, the first step now is to try and get a license for `ref=v{packageVersion}` and, if this fails, we fallback on master (just how I did before).

I'm not entirely sure about the `v` before `{packageVersion}`, but it seems quite common to have it here on GitHub so it should cover more cases than not using it.

I don't have problems in trying to get the license both with and without the `v` prefix by the way, but I think we should be fine already.

ST-Apps: Fine, let's summarize the cases we should handle:

1. https://github.com/coderpatros/dotnet-authenticated-httpclient/blob/master/LICENSE
2. https://raw.githubusercontent.com/coderpatros/dotnet-authenticated-httpclient/master/LICENSE
3. https://github.com/coderpatros/dotnet-authenticated-httpclient/blob/v1.3.0/LICENSE
4. https://raw.githubusercontent.com/coderpatros/dotnet-authenticated-httpclient/v1.3.0/LICENSE

This means that we could go with a regex like this one (added a capture group for `ref`):

`https?\:\/\/((github\.com)|(raw\.githubusercontent\.com))\/(?<repositoryId>[^\/]+\/[^\/]+)\/(blob\/)?(?<ref>.*)\/LICENSE.*$`

A possible flow could be this one:

1. Check license for given `ref` if `ref != "master"`
2. Check license for package version returned from NuGet (as `v{packageVersion}` and/or `{packageVersion}`)
3. Check license for `ref = "master"` as fallback

While steps 1 and 2 can be swapped, I'd say that `master` should always be the fallback because developers usually don't update the link to license file.

I'm not sure if I can work on this today, I'm afraid that I won't be available to make the change before tomorrow at 6pm GMT+1, so feel free to start writing some of the tests if you wish :)

coderpatros: @ST-Apps I've made some tweaks in PR #41 

There should always be a ref in the url (as far as I'm aware), so I've added that to the regex and always use it. Covers `master` and other branch refs as well as tags like `v1.0.0` and `1.0.0`.

I also split out github.com and raw.githubusercontent.com regex checks to make it a bit easier to read.

I don't think a licence should be resolved using anything other than the specified licence url. The chances of returning incorrect licence information is low. But I think it's better to err on the side of caution to avoid reporting an incorrect licence.

PR #65, opened 2020-01-29
Bumped versions to support .NET Core 3.1

karolswdev:
I do not have knowledge of your CI/CD pipeline so that it pushes it up to dotnet install tool will actually be downloading the 3.1. version. 3.1 is the de-facto version MS wants everyone to use.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -2,6 +2,10 @@ name: .NET Core CI
 
 on: [push, pull_request]
 
+  strategy:
+    matrix:
+      dotnet-version: ['2.1.803', '3.1.101']
+
 jobs:
   test:
     name: Test on ${{ matrix.os }}
@@ -15,6 +19,6 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.100
+        dotnet-version: ${{ matrix.dotnet-version }}
     - name: Tests
       run: dotnet test
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -8,7 +8,7 @@ on: [push, pull_request]
 
 jobs:
   test:
-    name: Test on ${{ matrix.os }}
+    name: Test ${{ matrix.dotnet-version }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -15,6 +15,6 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 2.2.108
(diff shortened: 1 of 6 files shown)


coderpatros:
I agree with you @stevespringett however I think 2.2 should be dropped back to 2.1 in the GitHub action. (2.1 is already the target framework in the project file)

2.2 is already EOL. 2.1 however is a LTS release like 3.1.

karolswdev:
Thanks @coderpatros and @stevespringett for reacting to this PR.

Good approach. I also think that 2.1 should be used instead of 2.2. 

Could you tell me how do you actually push it up as a dotnet tool? I wonder how to have two versions of this tool running simultaneously - for people who use .NET core SDK 2.1 & 3.1. Is dotnet smart enough to download the right version?

I'll make the changes once I get more clarity how I can support the above ^

coderpatros:
Steve packages and publishes it. The dotnet tools are basically just console apps bundled up as a nuget package.

So the project will need to be updated to multi-target the frameworks. These should help...
https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks
https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks

karolswdev:
@stevespringett - I went ahead and updated this PR with your requested changes.
@coderpatros - thanks for your insights!

karolswdev:
@stevespringett - apologies for the bump in advance. Do you have time and availability to process this PR in the coming days? The organization in which I work for is waiting for this to be supported in 3.X to complete our pipeline. No stress - if you are unable to get to this in the next couple of days we will work with a forked version of this repository in the interim.

stevespringett:
@coderpatros Any objections for merge?

.github/workflows/dotnetcore.yml
```diff
@@ -2,6 +2,10 @@ name: .NET Core CI
 
 on: [push, pull_request]
 
+  strategy:
+    matrix:
+      dotnet-version: ['2.1.803', '3.1.101']
+
 jobs:
   test:
     name: Test on ${{ matrix.os }}
```
coderpatros: @karolswdev can you change the name from `Test on ${{ matrix.os }}` to something like `Test ${{ matrix.dotnet-version }} on ${{ matrix.os }}`

coderpatros: Besides this little tweak it LGTM @stevespringett

karolswdev: Hi @coderpatros - I just created 88ff964 that addresses this, thanks

coderpatros:
@stevespringett just pinging you in case you didn't get pinged from the change @karolswdev made. Should be good to go.

coderpatros:
Oh, and please ping me back @stevespringett when this is merged.

#42 is ready for merge but there will be some minor merge conflicts with this pr that I want to resolve.

coderpatros:
@karolswdev and next time probably better to have the commit message as something like "Include .net version in github test action name" instead of "Addressing @coderpatros comments"

PR #90, opened 2020-03-07
Run DevSkim as a check for PRs

coderpatros:


coderpatros:
@stevespringett what do you think about adding [DevSkim](https://github.com/microsoft/devskim) checks for PRs?

I've been using it for a while and it seems ok. A bit of noise but does seem to add value.

stevespringett:
Never used it. As long as the tool provides accurate, actionable, and quality results, then sure. But it false positives are going to get in the way, then I’d rather not.

coderpatros:
I’ll leave it out @stevespringett

The checks it does are easy enough to pick up during code reviews.

The Roslyn analysers are much more accurate, pick up quality issues that are less obvious, and the rule set is much easier to fine tune to reduce noise. They also run automatically as part of the build.

Issue #122, opened 2020-04-27
Add support for JSON output


coderpatros:
In preparation of v1.2 of the spec we need to support JSON output.

coderpatros:
"preview" support available in 0.10.0, official support in 0.10.2

PR #124, opened 2020-04-28
Add initial preview support for JSON output

coderpatros:
Initial work for #122 @stevespringett I've only manually tested the JSON output so far.

But thought it might be good to get this out in the wild with JSON output a "preview" feature. What do you think?


stevespringett:
Fantastic. Lets do it

PR #178, opened 2020-07-03
Deterministic builds

coderpatros:


coderpatros:
There's still too much missing functionality from the .NET tooling.

PR #181, opened 2020-07-09
Add Artifactory compatibility to HttpClientExtensions

rocallaghan-deluxe:
Artifactory tends to return 405 errors when a file can't be found,
resulting in failures if using a private NuGet repo hosted on
Artifactory for the URL parameter.

### DIFF
--- a/CycloneDX.Core/Extensions/HttpClientExtensions.cs
+++ b/CycloneDX.Core/Extensions/HttpClientExtensions.cs
@@ -37,6 +37,10 @@ namespace CycloneDX.Extensions
             response = await httpClient.GetAsync(url).ConfigureAwait(false);
             
             if (response.StatusCode == System.Net.HttpStatusCode.NotFound) return null;
+
+            // JFrog's Artifactory tends to return 405 errors instead of 404
+            // errors when something can't be found.
+            if (response.StatusCode == System.Net.HttpStatusCode.MethodNotAllowed) return null;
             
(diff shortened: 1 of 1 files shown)


coderpatros:
@rocallaghan-deluxe have you reported this issue to JFrog?

I do want to support Artifactory. But don't like the idea of putting a vendor specific workaround in place like this.

rocallaghan-deluxe:
@coderpatros  I'll reach out to them, but I expect it's intentional - getting a file in Artifactory is technically an API call. The file doesn't exist, so a GET is invalid, but a PUT could be used to place a file there.

The main issue is probably more that other projects in the same solution file are being picked up as dependencies. They never get uploaded, so they'll never be found. The issue I'm having could also be resolved by preventing projects from being added to the `packages` HashSet in `ProjectAssetsFileService.GetNugetPackages(...)`. It's a fairly simple change (just checking `library.Type` and excluding `project` values), if that would be more acceptable then accepting 405 messages. Proper NuGet packages work with Artifactory as-is.

coderpatros:
@rocallaghan-deluxe yeah, I guess that is a valid interpretation of the status code.

coderpatros:
@rocallaghan-deluxe this is available in v0.11.1

Thanks for doing the leg work on this one. I know it's just a one liner. But you've saved us a fair bit of time by identifying the cause and contributing this PR.

PR #251, opened 2020-10-10
Customized build with central settings

Bertk:
- add Directory.Build.props and Directory.Build.targets for global build setting (see https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019)
- introduce [global.json](https://docs.microsoft.com/de-de/dotnet/core/tools/global-json?tabs=netcore3x)
- use [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning) for nuget and assembly versioning

ToDo:
- add `.editorconfig` file with configuration for tabs, etc. 
- add [GitHub Action for Nerdbank.GitVersioning](https://github.com/marketplace/actions/nerdbank-gitversioning) in build pipeline


coderpatros:
@Bertk before you put any more effort into this PR, I don't think it will be merged as it currently is. There's a lot going on here that is unrelated. And some stuff, that wouldn't be accepted individually.

One thing that is unlikely to be accepted is changing the approach to versioning. We've chosen, where possible, to keep a consistent approach to versioning across implementations. We have to jump back and forth between different ecosystems and consistency in our versioning and release process makes that easier.

The .editorconfig addition I'd prefer in a PR by itself. And the same goes for Directory.Build.Targets changes.

We also currently support .NET Core 2.1 and 3.1 as the current LTS versions. I haven't had a need for global.json yet, but I don't think it supports specifying two different major.minor version combinations?

We're definitely keen to receive PRs. But if the scope of the change is outside of an existing issue, bug fix or minor improvement it's probably best to reach out first. Either by creating a GitHub issue to discuss an idea with us or even jumping on the #dotnet channel of the CycloneDX Slack (invite link is available on the CycloneDX website).

Bertk:
@coderpatros  Thank you for the hints. I will remove the global.json, gitversioning and the roslyn analyzer section in Directory.Build.props and create new PRs.

PR #355, opened 2021-03-07
Update CODEOWNERS

jkowalleck:
fixed team name
and added some comments


jkowalleck:
@coderpatros 

there was a slight change in the codeowners. team name had no organization.
this is why this PR is probably not assigned to you automatically. so i added you as a reviewer manually.

PR #376, opened 2021-04-22
Add support for component hashes

jakrivan:
Component hashes populated from .nupkg.sha512 or calculated from .nupkg content


jakrivan:
There is something weird in the failing test runs. I'll have a closer look next week, but if anybody has any idea why the following can be happening, than I'd definitely appreciate it.

The tests are failing for `System.Runtime.InteropServices.RuntimeInformation@4.0.0` - only on windows platform, succeeding on other platforms. Nugets are not platform specific - why would the hash be different on different platforms?
When I manully download the package and calculate the hash - it's exactly same as in the resource file:

```
$response = Invoke-WebRequest -Uri "https://api.nuget.org/v3-flatcontainer/System.Runtime.InteropServices.RuntimeInformation/4.0.0/System.Runtime.InteropServices.RuntimeInformation.4.0.0.nupkg"
[System.BitConverter]::ToString( $hasher.ComputeHash($response.Content) ).Replace("-","")
```

result: `462D35E66CBDD21DC007F06C6EF129AB57E810FA0F0416BD2FC6FB7EED55138780D4D31E31EE6267A82E2E3A1607E5C642BD6EFEB130B57A1BAA87E3141B0080`

while the windows test run on windows platform generates something else:

```
Expected: ···ash alg=\"SHA-512\">462D35E66CBDD21DC007F06C6EF129AB57E810FA0···
Actual:   ···ash alg=\"SHA-512\">8563E1271738E774426BC676F4EF751267C679920···
```

How exactly is the windos container prepared? Id' liketo reproduce same conditions as in the CI run.

coderpatros:
Yeah, I tried running the tests locally and had the same issue. Strange, because I’m on Linux and those tests pass on the hosted Linux runner. 

There are some oddities as you get closer to the lower levels of .NET. But I haven’t tried picking it apart yet.

jakrivan:
I'll be happy to deep dive into it, but I need a repro. Can I somewhere grab the docker image used for the windows hosted runner? Or any other way of simulating CI run locally? (sorry I'm new to github Actions)

Local failing tests might be a result of nugets from other then nuget.org repositories, as those might be adding their signatures (`dotnet nuget list source` for list of sources, In my case I needed to disable other sources, clear the cached nugets `dotnet nuget locals all --clear` and then all was good... except that single `System.Runtime.InteropServices.RuntimeInformation@4.0.0` package).

coderpatros:
I wonder if those packages are being pulled in from the SDK? I have the offending ones in my local nuget fallback directory.

coderpatros:
And don’t know how to repro the hosted runners locally. I think the windows one is actually spinning up a VM in Azure under the hood.

jakrivan:
The fallback dir is a very valuable observation! Thanks for that!
I added force disabling of usage of the fallback (as we should only pull from common reference source to ensure consistency) - that should catch that case.
If there still will be mismatch between platforms than I'll try to get a repro localy

coderpatros:
I don't think we should ignore the nuget fallback folder. Typically the package will be pulled from there, not nuget.org. So we should be using the hash of the package from that location.

Although, being able to provide that as an option might be handy. Even if it's just internally for use during tests.

The integration tests are pretty over the top. I never got around to refactoring them after building out the unit tests and refactoring a bunch of stuff into the CycloneDX .NET library. So I'm not adverse to dropping some problematic tests either.

jakrivan:
I was thinking that the local nuget fallback might make the generated SBOMs less transferable between platforms... but it seems to affect just that on particular .nuget and getting rid of local nugets fallback folder doesn't solve that either.

I'll just reverse those changes and will resolve the unit test different way (worst case removing the VB unit test).

Opt-in skipping of local nuget fallback folder can be left for separate PR :)

jakrivan:
Ok - so I've reverted my changes affecting the usage of fallback folder (I may have create a separate PR for opt-in to skipping it).
I've as well forced the VB unit tests to use newer version of System.Runtime.InteropServices.RuntimeInformation, which doesn't have the issue with hash mismatch.

All the test runs are now passing on my fork Action pipeline.

Thank you for the patience with this PR :-)

jakrivan:
@coderpatros Is there anything how I can improve this PR? I'm happy to iterate. Plus we are looking forward to use this new version in our build pipeline.

coderpatros:
It all looks good to me. Just been busy :)

Big thanks for the PR!

coderpatros:
Now available in version 1.5.0.

jakrivan:
Awesome - thanks Patrick

davideicardi:
I have just created a flag to disable hash computation, because if this is not necessary can be quite time consuming ...

See #456 . What do you think?

Issue #415, opened 2021-08-17
Components of test projects should not be of scope required


patspaeth:
I would assume that the scope required is only for packages that will be required to run the application.
When I do not exclude test projects all components of these test projects are marked as required, but I would expect that they are marked as exclude or similar

coderpatros:
Yeah, that should be the behaviour. The check is missing in the ProjectAssetsFileService.GetNugetPackages() method.

patspaeth:
The question is: Shall all transitive dependencies be gathered or only the top level packages?
For me, the top level packages would be enough, what do you think?

coderpatros:
Resolved in v1.6.2

coderpatros:
Thanks for the PR @patspaeth

PR #417, opened 2021-08-18
Packages from Testprojects will marked as scope="excluded"

patspaeth:
Related to issue #415 

Signed-off-by: Patrick Spaeth <patrick.spaeth@t-online.de>


patspaeth:
@coderpatros I am getting an error in the pipeline for the test: GitHubApiRateLimitExceeded (exitcode = 6)
On my local PC it is working, do you have an idea?

coderpatros:
> @coderpatros I am getting an error in the pipeline for the test: GitHubApiRateLimitExceeded (exitcode = 6)
> On my local PC it is working, do you have an idea?

Might just be an unlucky run. This was always a possibility. But I have never worried about it because I haven't hit the limit yet from the hosted runners. I'm re-running the checks.

patspaeth:
It seems very deterministic - macOS and my newly added tests. Unfortunately I cannot retest it here (no mac-os)
Can we provide some github credentials for the runners?

PR #434, opened 2021-10-01
Version 3 working branch

coderpatros:
- [x] dependency graph support
- [x] drop support for .NET Core 2.1 runtime
- [ ] migrate to `System.CommandLine` from `McMaster.Extensions.CommandLineUtils`
- [ ] refactor Program.cs
- [ ] support specifying output filename
- [ ] support no parameter BOM generation
- [ ] exclude test and development dependencies by default
- [x] remove deprecated parameters
- [ ] make GitHub license resolution opt-in
- [ ] add protobuf output support
- [ ] detect output format from output filename if provided
- [ ] read github credentials from environment variables
- [ ] read metadata automatically from base BOM .cdx/base.[xml|json|bin]

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


mtsfoni:
Doesn't look like there is anything in that branch that we don't already have one way or the other in master.
I close this, but added the open points from the list as issues for the future.

PR #448, opened 2021-11-04
Allow to disable package restore

### LINKED ISSUE #447: Allow to disable package restore
Is really necessary to always execute `dotnet restore` before generating BOM?

If I'm inside a pipeline that already do several steps and already restore packages should I really need to restore package again?

Maybe we can have a flag to disable it? Can I try to implement it in a new PR?

The reason why I prefer to skip restore is because otherwise I need to configure nuget credentials (and can potentially restore different packages?). I run cyclonedx inside docker, and passing credentials is currently not supported or not easy I suspect.

davideicardi:
Fix #447 

Signed-off-by: Davide Icardi <davide.icardi@gmail.com>


davideicardi:
I have added a `DisablePackageRestore` property, that is passed to the `ProjectFileService`. When `true` package restore is skipped.

Also added a test to verify that `Restore` was never called when `DisablePackageRestore=true`.

coderpatros:
@davideicardi thanks for the PR!

PR #456, opened 2021-11-26
add flag to disable hash computation for packages

davideicardi:
Add a flag (`disable-hash-computation`) to disable the computation of hash for components.
Main reason for that is to speed up the BOM generation when we don't need hash for components.

I maintaned the hash only in case it was already available as `.nupkg.sha512`, because my assumption is that in this case the performance penality is small.

### DIFF
--- a/CycloneDX.Tests/NugetServiceTests.cs
+++ b/CycloneDX.Tests/NugetServiceTests.cs
@@ -171,7 +171,7 @@ namespace CycloneDX.Tests
 
             var component = await nugetService.GetComponentAsync("testpackage", "1.0.0", Component.ComponentScope.Required).ConfigureAwait(false);
 
-            Assert.Equal(null, component.Hashes);
+            Assert.Null(component.Hashes);
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


davideicardi:
@coderpatros  What do you think? Thank you

davideicardi:
@coderpatros I see that one check failed due to .NET 2.1 issues, but are not caused by my changes.
If you want anyway I can try to fix it in this or in another PR.

davideicardi:
@stevespringett  Do you think we can merge this PR? Do you see problems? Thank you!

stevespringett:
Hi @davideicardi. I am not well versed in the codebase, so I defer all decisions to @coderpatros who is the maintainer of this repo.

coderpatros (approved):

coderpatros:
@davideicardi thanks for PR!

Apologies it took a while to get to it.

davideicardi:
@coderpatros  Thank you! Do you think it is possible to generate a new release?

coderpatros:
Oh, I thought I had. Maybe I forgot. I've been busy. But have some time this week. I'll get it released.

PR #515, opened 2022-03-24
Add flexibility to Regex and add main lookup

killi199:
Adds license support for the non-American spelling. And adds support for the main branch instead of only the master.

### DIFF
--- a/CycloneDX/Services/GithubService.cs
+++ b/CycloneDX/Services/GithubService.cs
@@ -67,8 +67,8 @@ namespace CycloneDX.Services
         private readonly HttpClient _httpClient;
         private readonly List<Regex> _githubRepositoryRegexes = new List<Regex>
         {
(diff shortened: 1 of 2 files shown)


CycloneDX/Services/GithubService.cs
```diff
@@ -68,8 +68,8 @@ public class GithubService : IGithubService
         private HttpClient _httpClient;
         private List<Regex> _githubRepositoryRegexes = new List<Regex>
         {
-            new Regex(@"^https?\:\/\/github\.com\/(?<repositoryId>[^\/]+\/[^\/]+)\/((blob)|(raw))\/(?<refSpec>[^\/]+)\/[Ll][Ii][Cc][Ee][Nn][Ss][Ee]((\.|-)((md)|([Tt][Xx][Tt])|([Mm][Ii][Tt])|([Bb][Ss][Dd])))?$"),
```
rilehudd, line 71: I know this is just an update to an existing regex.
But what about using RegexOptions.IgnoreCase to simplify?

coderpatros (approved):

coderpatros:
Thanks for the PR @killi199

PR #516, opened 2022-03-25
Using package version for which a dependency version ranges resolves to. Improving top level component detection.

rilehudd:
1. There were some cases where a framework dependency lookup for a version range resulted in failure to find the resolved dependency (resulting in a KeyNotFoundException).
* Fixed this by resolving ranges to components that have a specific version that satisfys the version range. 

2. If a package was both a top level reference and a transitive reference, the package was not added to the top level component's dependency.
* Utilized the dotnet command to get a project's direct package dependencies (with non-transitive being the fallback determination).

Added 2 tests for increased coverage relating to these issues: [TestResults.zip](https://github.com/CycloneDX/cyclonedx-dotnet/files/8353260/TestResults.zip)


CycloneDX/Program.cs
```diff
@@ -305,10 +308,12 @@ public static async Task<int> Main(string[] args)
             {
                 return (int)ExitCode.GitHubLicenseResolutionFailed;
             }
+
+            var directPackageDependencies = packages.Where(p => p.IsDirectReference).Select(p => packageToComponent[p].BomRef).ToHashSet();
             // now we loop through all the dependencies to check which are direct
             foreach (var dep in dependencies)
             {
-                if (!transitiveDepencies.Contains(dep.Ref))
+                if (directPackageDependencies.Contains((dep.Ref)) || !transitiveDepencies.Contains(dep.Ref))
```
rilehudd, line 316: Seems that I should have probably removed the old condition instead of combining it. 
This causes transitive framework dependencies to still be considered top level dependencies.

rilehudd: Updated to prioritize directPackageDependencies when filled over list transitive dependencies.

PR #531, opened 2022-05-07
Updated dependencies and nuget service, plus refactoring

verzada:
Upgraded all of the dependencies.
Created a new nugetservice which uses nuget.protocol instead of httpclient
Refactored code, separated interfaces from services
Removed support for .net2.1 and .net5.0 due to lts ending.


verzada:
I've upgraded all of the packages in my fork.
As well as removed httpclient and instead using nuget.protocol package to handle nuget requests, plus fixed an issue or two.

Not sure why httpclient was used to begin with, but I guess I’ll learn about it if it’s an issue.

Got a gut feeling I've brute forced the changes, so it might not be viable to introduce my fork. However, I've solved an issue with nuget api keys for especially private repositories.

Hope I didn't do too much of a mess of it @stevespringett

verzada:
Seems like a few nuget package references has been muddled up.

PR #535, opened 2022-05-10
Output help text when invalid parameters are entered

### LINKED ISSUE #239: Should output help text if no paramaeters are provided
Current behaviour outputs `A path is required` if no parameters are supplied.

It would be more helpful to default to the full help text if no parameters are provided.

RodneyRichardson:
Calls `ShowHelp` with appropriate error message when commandline arguments are invalid:

- when path is not supplied
- when outputDirectory is not supplied
- when only partial github credentials are supplied

Fixes #239


VinodAnandan:
Slack invite link - https://cyclonedx.org/slack/invite

Issue #489, opened 2022-02-07
License resolution failing to get correct license


alitheg:
Please tell me if this is actually a problem importing into DependencyTrack, but it feels like an issue generating the BOM. I've got several dependencies (for example, Microsoft.Azure.Cosmos and System.Buffers) which don't show a license when imported into DependenctTrack. I've looked at the bom.xml generated, and in these cases the license is set to a URL. In the examples above, it's:

- Microsoft.Azure.Cosmos
-- https://aka.ms/netcoregaeula
- System.Buffers
-- https://github.com/dotnet/corefx/blob/master/LICENSE.TXT

Both of these URLs resolve into a MIT license file if you follow them, so shouldn't the BOM be generated showing that as the license instead of just the URL?

rstmm:
I can add some here, e.g.
Google.Api.CommonProtos 2.2.0
Microsoft.ApplicationInsights.Kubernetes 2.0.2
 -> both don't resolve but add  URL https://aka.ms/deprecateLicenseUrl. in NuGet and Github  a license is defined properly it seems

rstmm:
Ok the license detection of the nuget tool seems to be way worse than the npm tools. But I did dig a bit deeper and this seems to be mostly because of information missing in NuGet repo.
Compare
https://api.nuget.org/v3-flatcontainer/Keycloak.Net/1.0.18/Keycloak.Net.nuspec
with
https://api.nuget.org/v3-flatcontainer/AutoMapper/11.0.1/AutoMapper.nuspec

AutoMapper contains a license URL AND a license expression, hence is correctly identified in BOM. Keycloak is only returning a license URL, but lacks the expression, and then only the URL is contained in BOM. Also GitHub seems not to be requested in this case as far as I can see, cause there the license is properly defined. Probably as soon as *anything* regarding license is returned, that will be taken and no further investigation done via github etc. 

And my examples with the deprecateLicenseUrl also are not a tool failure, since the NuGet API is returning exactly this information: https://api.nuget.org/v3-flatcontainer/Google.Api.CommonProtos/2.2.0/Google.Api.CommonProtos.nuspec

I added this to 3 projects now and licenses are not identified for 50-80% in these projects, which means its pretty much missing most of the licenses (talking about 120-180 dependencies per project). Any ideas how to improve this?

RodneyRichardson:
Would it be worth downloading and parsing the license text, such as done by [licensee](https://github.com/licensee/licensee)?

[...]

PR #553, opened 2022-05-20
Added repository and commit fallback for non-GitHub URLs in licenseUrl field

### LINKED ISSUE #432: Github license resolution fails for repos not using "master"
As lots of repos have now moved away from calling the default branch "master", this library fails Github license resolution for those libraries now. Perhaps the support ticket referenced in the code will get fixed, but if not it might be worth failing once and trying "main" too.

ecaisse:
Fixes #432 and #489.

Added `--enable-github-licenses-non-default-branch` switch to opt-in to fetch license information for branches other than `master` or `main`. This ensures that current behavior doesn't change.
Added fallback to using the repository URL and commit ref (if both present) to query the GitHub API for a license if the licenseUrl is not a GitHub URL. This requires the usage of `--enable-github-licenses-non-default-branch` since the commit ref is always in hash form.
Added `main` as a default branch alongside `master`. Breaking change, but minimal impact.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -64,7 +64,7 @@ jobs:
         run: dotnet test --framework ${{ matrix.framework }}  --collect:"XPlat Code Coverage;Format=cobertura" --results-directory "TestResults"
       # see https://github.com/danielpalme/ReportGenerator/blob/main/.github/workflows/ci.yml
       - name: ReportGenerator
-        uses: danielpalme/ReportGenerator-GitHub-Action@5
+        uses: danielpalme/ReportGenerator-GitHub-Action@v5
         with:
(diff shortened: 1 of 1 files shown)


coderpatros (approved):

BlythMeister:
the github api serves the same licence regardless of the branch ref.

BlythMeister:
https://api.github.com/repos/BlythMeister/LicenceChangeTest/license?ref=f99a1057598f39b631667e81edb5d0610c2bc4cb for example.  On this commit, the licence should be GNU
but on master it was updated.

It doesn't matter what ref you pass, you always get the licence for the default branch.
Their docs (https://docs.github.com/en/rest/licenses?apiVersion=2022-11-28#get-the-license-for-a-repository) actually say there is no refspec argument. :(

I have conversed with github support over this, and this is "intended behaviour"

PR #615, opened 2022-11-15
Additional Lookups For Licence

BlythMeister:
Have seen situations whereby the licence URL isn't github (often when it's a file) and using the repository (or project) url then locates the licence

Logic now locates licence based on this order

* Expression in the package (this is where Nuget has found the licence itself)
* Licence URL node
* Repository node with commit SHA
* Repository node without ref
* Project URL

These checks all assume that the URLs are github, the github service will only check valid URLs

Secondary updates to the GitHub service:

* Use the actual licence URL for the located licence (sometimes this differs from the input URL)
* Allow a licence lookup with a base repository URL - use the current default branch licence

### DIFF
--- a/CycloneDX.Tests/GithubServiceTests.cs
+++ b/CycloneDX.Tests/GithubServiceTests.cs
@@ -402,5 +402,28 @@ namespace CycloneDX.Tests
 
             mockHttp.VerifyNoOutstandingExpectation();
         }
+
+        [Fact]
+        public async Task GitLicence_DifferentHtmlUrl()
+        {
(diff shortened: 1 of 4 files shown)


coderpatros (approved):

BlythMeister:
giphy

coderpatros:
Ha ha, I was just coming back to say thanks for another great contribution @BlythMeister!

The release process has just been triggered. It should be published soon.

BlythMeister:
Lovely.  Loving working on this (and dependency track) such a nice framework to work with and a very friendly community!

PR #625, opened 2022-11-25
GitHub Performance & Edge case improvements

BlythMeister:
* If getting licence and repository URL ends ".git" and you get a NOTFOUND (404) try again without ".git"
* Add support for the URL being a SSH URL (git@github.com/owner/repo.git)
* Add Cache of URL to licence so if same URL requested, return from the previous result rather than getting again from GitHub

### DIFF
--- a/CycloneDX.Tests/GithubServiceTests.cs
+++ b/CycloneDX.Tests/GithubServiceTests.cs
@@ -15,17 +15,18 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (c) OWASP Foundation. All Rights Reserved.
 
+using System.Net;
 using System.Threading.Tasks;
-using Xunit;
-using RichardSzalay.MockHttp;
 using CycloneDX.Services;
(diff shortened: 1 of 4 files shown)


BlythMeister:
@coderpatros the warning is to do with code analysers.  I _think_ you no longer need the nuget package.
In our company codebase we stopped adding it for net5+ projects, but since Cyclone also has a 3.1 i'm not sure if it should be removed or not.

BlythMeister:
@coderpatros have put in to skip the code analyser out of date warning.

I think this change might conflict with #626

BlythMeister:
@coderpatros i'm seeing issues with our builds because of rate limiting (even though we have the user creds set) so think this and #626 are required :(

Any chance of getting these in and released?

coderpatros (approved):

BlythMeister:
@coderpatros thanks, not got permission to solve conflicts on the other PR ☹️

PR #640, opened 2022-12-15
Fix bug when nuget package has a multiple condition licence expression

BlythMeister:
Example package with multiple licence expression: https://www.nuget.org/packages/rabbitmq.client/

Took the licence expression and included in test.


coderpatros (approved):

BlythMeister:
Thanks @coderpatros

coderpatros:
> Thanks @coderpatros 

Thank you for another contribution @BlythMeister

PR #647, opened 2023-01-16
Feature/dotnet 7

MattBasson:
Adding .net 7 package references and changing docker base image to 7


ErikThorsell:
Hey, 

What needs to be done here for this to proceed? :)

MattBasson:
Awating review and merge from CycloneDX in particular @coderpatros

pombredanne:
LGTM to me... but this is missing the tests! do you mind adding 7 to https://github.com/CycloneDX/cyclonedx-dotnet/blob/master/.github/workflows/dotnetcore.yml ?

MattBasson:
> LGTM to me... but this is missing the tests! do you mind adding 7 to https://github.com/CycloneDX/cyclonedx-dotnet/blob/master/.github/workflows/dotnetcore.yml ?

Have added .net 7 to the github workflows as suggested, let  me know if its ok @pombredanne

pombredanne:
@MattBasson all green :green_heart: ... but I am not sure why there seem to be some error in the logs such as https://github.com/CycloneDX/cyclonedx-dotnet/actions/runs/4128554332/jobs/7133528411#step:6:86 ? this may not be your making at all since there are similar failures in other .net versions? Also I am not sure, but the tests for net6 seem to run on net7 ... is this something you mind to double check?

Last, I am not sure what Patrick may want, but if I were you, I would squash the commits in a single one with a clean commit message.

MattBasson:
> @MattBasson all green 💚 ... but I am not sure why there seem to be some error in the logs such as https://github.com/CycloneDX/cyclonedx-dotnet/actions/runs/4128554332/jobs/7133528411#step:6:86 ? this may not be your making at all since there are similar failures in other .net versions? Also I am not sure, but the tests for net6 seem to run on net7 ... is this something you mind to double check?
> 
> Last, I am not sure what Patrick may want, but if I were you, I would squash the commits in a single one with a clean commit message.

Hi there, I am not seeing any errors at the link you sent.
image

As for the tests all running on dotnet 7 prior to this all the tests were running on dotnet 6 i believe based upon your release.yml file which I have also included changes to  include dotnet 7 and also set the tests to run on dotnet 7.
Thes should still be backward compatible with older versions of dotnet like dotnet6 was when it was running older versions .
see line 46

image

If this is incorrect I can revert it back to 6.

As for the squash commit,  I dont appear to have access to do this from my end, I think this is something the approver who will have the privelage to make the choice as to how this gets merged into your master branch.

Hope this helps and let me kno if there is anything more I can help with.

:)

pombredanne:
> Hi there, I am not seeing any errors at the link you sent.

I think there could be an error because of this output at https://github.com/CycloneDX/cyclonedx-dotnet/actions/runs/4128554332/jobs/7133528411#step:6:86
```
A .NET Core global tool which creates CycloneDX Software Bill-of-Materials (SBOM) from .NET projects.

Usage: dotnet cyclonedx [options] <path>

Arguments:
  path                                                                   The
                                                                         path to
                                                                         a .sln,
<.... clipped for brevity>

  -?|-h|--help                                                           Show
                                                                         help
                                                                         information.

The output directory is required
```

Here `The output directory is required` likely means that there is a test that may not work. It could also be that the CLI does NOT return a proper error code as it should? Again this is all prospective but worth digging even if this is not your making.


> As for the tests all running on dotnet 7 prior to this all the tests were running on dotnet 6 i believe based upon your release.yml file which I have also included changes to include dotnet 7 and also set the tests to run on dotnet 7.
> Thes should still be backward compatible with older versions of dotnet like dotnet6 was when it was running older versions 

Is there a way to modify the tests to ensure that the tests are running explicitly and separated on the three framework versions supported here? Here again the issue may not have been yours in the first place, but worthy of digging and fixing now IMHO. 


> As for the squash commit, I dont appear to have access to do this from my end, I think this is something the approver who will have the privelage to make the choice as to how this gets merged into your master branch.

You have all the powers to squash and force push to your branch alright, this is not something that should be done at merge time IMHO. Squash in your clone crafting a proper message and then you can force push your branch to keep a single clean commit.

.devcontainer/Dockerfile
```diff
@@ -9,4 +9,8 @@ RUN apt-get update \
     && apt-get update \
     && apt-get install -y dotnet-sdk-2.1 \
     && apt-get install -y dotnet-sdk-3.1 \
+    && wget https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb \
```
jkowalleck, line 12: :exclamation:  this image is based on Ubuntu 20.04
and you are installing ubuntu 22.04 packages? "It works" is not an excuse! Please get this sorted out.

Furthermore, I'd suggest to put some version on the multiple packages-microsoft-prod.deb that are stored on disk, here.

.github/workflows/dotnetcore.yml
```diff
@@ -25,6 +25,10 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: '6.x'
+      - name: Setup dotnet 7
```
jkowalleck, line 28: :wrench:  please take the change and change this to the 
"multi version syntax" as described here https://github.com/actions/setup-dotnet#usage

```yaml
steps:
- uses: actions/checkout@v3
- name: Setup dotnet
  uses: actions/setup-dotnet@v3
  with:
    dotnet-version: | 
      3.1
      3.1.x
      7.x
```

jkowalleck (changes requested):
i have some critical remarks:

.devcontainer/Dockerfile
MattBasson: Happy to sort @jkowalleck , would the sort be changing Ubuntu to 22.04 and installing all versions from the 22.04 microsoft source?
as dotnet 7 is not available on the 20.04 microsoft source.

.github/workflows/dotnetcore.yml
MattBasson: @JanKrivanek  I can change this too  would this also apply to .github/workflows/release.yml
image

.devcontainer/Dockerfile
MattBasson: @jkowalleck  Also to mention NET Core 3.1 or 2.0 are not supported on Ubuntu 22.04 Because the distro only supports OpenSSL 3, so not sure how you would lay this out in your solution?
1 devcontainer per Ubuntu version?

jkowalleck: I leave the decision to @coderpatros.
I do not know much about the specifics of .net.

.github/workflows/dotnetcore.yml
MattBasson: @jkowalleck  Have made these changes to the yml files

jkowalleck: i do not know if the release process needs all .net versions installed to do the job. 
you decide.

.devcontainer/Dockerfile
MattBasson: @jkowalleck  have created a seprated dockerfile that runs on Ubuntu 22.04 with dotnet 7 and the older config i have left on  its own docker file, and defaulted the devcontainer config to the Ubuntu 22.04 docker file.
Let me know if this works

MattBasson:
> > Hi there, I am not seeing any errors at the link you sent.
> 
> I think there could be an error because of this output at https://github.com/CycloneDX/cyclonedx-dotnet/actions/runs/4128554332/jobs/7133528411#step:6:86
> 
> ```
> A .NET Core global tool which creates CycloneDX Software Bill-of-Materials (SBOM) from .NET projects.
> 
> Usage: dotnet cyclonedx [options] <path>
> 
> Arguments:
>   path                                                                   The
>                                                                          path to
>                                                                          a .sln,
> <.... clipped for brevity>
> 
>   -?|-h|--help                                                           Show
>                                                                          help
>                                                                          information.
> 
> The output directory is required
> ```
> 
> Here `The output directory is required` likely means that there is a test that may not work. It could also be that the CLI does NOT return a proper error code as it should? Again this is all prospective but worth digging even if this is not your making.
> 
> > As for the tests all running on dotnet 7 prior to this all the tests were running on dotnet 6 i believe based upon your release.yml file which I have also included changes to include dotnet 7 and also set the tests to run on dotnet 7.
> > Thes should still be backward compatible with older versions of dotnet like dotnet6 was when it was running older versions
> 
> Is there a way to modify the tests to ensure that the tests are running explicitly and separated on the three framework versions supported here? Here again the issue may not have been yours in the first place, but worthy of digging and fixing now IMHO.
> 
> > As for the squash commit, I dont appear to have access to do this from my end, I think this is something the approver who will have the privelage to make the choice as to how this gets merged into your master branch.
> 
> You have all the powers to squash and force push to your branch alright, this is not something that should be done at merge time IMHO. Squash in your clone crafting a proper message and then you can force push your branch to keep a single clean commit.

A

.devcontainer/Dockerfile
jkowalleck: @coderpatros needs to decide. 
I just pointed out some oddities.

coderpatros: I think bumping the base image to 22.04 makes sense. It's the current LTS version. .NET Core 3.1 and 2.0 are not supported anymore. So I'm happy to do a major version bump and drop support for them.

MattBasson:
@coderpatros  @pombredanne  @jkowalleck  have squashed all commits, 
I have also removed .net core 3.1 testing and support and have updated the readme.
could you please inform me if there are any further requirments before this can be accepted?

Regards 

Matt

jkowalleck:
tests are passing.

@coderpatros could you review and maybe megre?

jkowalleck:
continues in #666

PR #672, opened 2023-04-03
use central package management and update dependencies

Bertk:
- Microsoft introduced CPM which simplifies nuget package handling  see also https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#transitive-pinning
- update all dependecies


coderpatros (approved):

coderpatros:
Thanks @Bertk

Issue #603, opened 2022-10-14
System.IO.InvalidDataException: Central Directory corrupt with 2.4.1


Philipp-Navis:
Hi,

using version 2.4.1 some of our builds are failing due to a `System.IO.InvalidDataException` exception. Those builds were working using version 2.3.0 and are still working if we fix the version to 2.3.0. We observed this exception in two different projects for different dependencies:
```
Retrieving GitHub license for repository nspec/NSpec and ref master
Unhandled exception. System.IO.InvalidDataException: Central Directory corrupt.
 ---> System.IO.IOException: An attempt was made to move the position before the beginning of the stream.
   at System.IO.MemoryStream.Seek(Int64 offset, SeekOrigin loc)
   at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
   --- End of inner exception stack trace ---
   at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
   at System.IO.Compression.ZipArchive.Init(Stream stream, ZipArchiveMode mode, Boolean leaveOpen)
   at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen, Encoding entryNameEncoding)
   at CycloneDX.Services.NugetV3Service.GetNuspec(String name, String version, String nuspecFilename, FindPackageByIdResource resource) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 315
   at CycloneDX.Services.NugetV3Service.GetNuspec(String name, String version, String nuspecFilename, FindPackageByIdResource resource) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 315
   at CycloneDX.Services.NugetV3Service.GetComponentAsync(String name, String version, Nullable`1 scope) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 231
   at CycloneDX.Services.NugetV3Service.GetComponentAsync(NugetPackage nugetPackage) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 322
[...]

PR #674, opened 2023-04-03
Normalize version string when locating files in the local cache

### LINKED ISSUE #673: Nuget packages with version x.y.z.0 are not located correctly in local cache
Nuget normalizes version strings according to 
https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#normalized-version-numbers. 
CycloneDX fails to locate cached packages with version x.y.z.0, which are stored in a folder named x.y.z . It will
try to load the files from the web.

If CycloneDX would normalize the version strings when locating the files in the local cache, loading from the remote
repo can be avoided.
Relates to #603 "System.IO.InvalidDataException: Central Directory corrupt with 2.4.1"

Arthur-van-Dongen:
Why: Nuget will normalize the version string according to 
https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#normalized-version-numbers, 
when creating the local cache. Thus, a version "1.0.0.0" will be stored in folder "1.0.0" and the nupkg 
and nupkg.sha512 files will be named accordingly.

How: This change adds a Normalize method to strip off the Revision number under certain conditions, before locating the files in the cache.
Fixes #673, relates to #603

### DIFF
--- a/CycloneDX.Tests/NugetV3ServiceTests.cs
+++ b/CycloneDX.Tests/NugetV3ServiceTests.cs
@@ -80,37 +80,52 @@ namespace CycloneDX.Tests
             Assert.Equal("testpackage", component.Name);
         }
 
-        [Fact]
-        public async Task GetComponent_FromCachedNugetHashFile_ReturnsComponentWithHash()
+        public static IEnumerable<object[]> VersionNormalization
         {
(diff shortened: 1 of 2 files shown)


coderpatros:
Thanks for the contribution @Arthur-van-Dongen

PR #675, opened 2023-04-04
fix: exclude developmentDependency

### LINKED ISSUE #665: developmentDependency not excluded from SBOM
In quite a number of repositories, we have a `Directory.Build.props` file looking like this:

```xml
<!--EXTERNAL_PROPERTIES: IsTestProject-->
<Project>

  <PropertyGroup>
    <CodeAnalysisRuleSet>..\sonarqube.api.ruleset</CodeAnalysisRuleSet>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="SonarAnalyzer.CSharp" Version="8.53.0.62665" developmentDependency="true" Condition="'$(IsTestProject)' != 'true'">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>

  <Target Name="DisableAnalyzersForTestProjects" BeforeTargets="CoreCompile" Condition="'$(IsTestProject)' == 'true'">
    <ItemGroup>
      <Analyzer Remove="@(Analyzer)"/>
    </ItemGroup>
  </Target>

</Project>
```

Note that we mark the `SonarAnalyzer.CSharp` reference as  `developmentDependency="true"`.

when we generate the SBOM, we run the command like:

```
dotnet CycloneDX -o $PublishPath $solutionPath -j -t -d -dpr -f $projectName.sbom.json -st $Classifier -sn $projectName -sv $PackageVersion 
```

We expected that because we specify the `-d` option, we won't see `SonarAnalyzer.CSharp` in the SBOM's anymore, but it still pops up. In addition to that, the nuspec file for SonarAnalyzer.CSharp also states that it is a development Dependency.

Why is this dependency still showing up in our SBOM's if we explicitly mark it to be excluded?

mieliespoor:
Performs a check on whether the dependency should be excluded and then skip over adding to the BOM

fixes #665

### DIFF
--- a/CycloneDX/Program.cs
+++ b/CycloneDX/Program.cs
@@ -197,7 +197,7 @@ namespace CycloneDX
                 HttpClient httpClient = new HttpClient(new HttpClientHandler {
                     AllowAutoRedirect = false
                 });
-                
+
                 if (!string.IsNullOrEmpty(githubBearerToken))
                 {
                     githubService = new GithubService(httpClient, githubBearerToken);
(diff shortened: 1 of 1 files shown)


Bertk:
Please request a review by @coderpatros and **signoff the PR** (see https://github.com/apps/dco).

I confirm the PR removes development dependencies and transitive dependencies. My sample uses
- AsyncFixer
- coverlet.collector
- SonarAnalyzer
- GitVersioning

PR #711, opened 2023-05-27
use System.CommandLine instead of McMaster.Extensions.CommandLineUtils

Bertk:
System.ComponentModel.Annotations V5.0.0 is deprecated (see also https://github.com/natemcmaster/CommandLineUtils/issues/533)

**BREAKING CHANGE**: commandline option names are updated and deprecated options are dropped.
- option name `--output` is used instead of former `--out`. This avoids conflict with **_out parameter modifier_** and is compliant to recommendations [Options as parameters](https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#options-as-parameters)
image

### DIFF
--- a/CycloneDX.Tests/ProgramTests.cs
+++ b/CycloneDX.Tests/ProgramTests.cs
@@ -31,19 +31,19 @@ namespace CycloneDX.Tests
     public class ProgramTests
     {
         [Fact]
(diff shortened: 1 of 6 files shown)


Bertk:
@coderpatros  
Hi, could you please send me a hint how this security check problem should be resolved. This is located in git history 🤨

image

PR #716, opened 2023-06-03
use System.CommandLine instead of McMaster.Extensions.CommandLineUtils

Bertk:
System.ComponentModel.Annotations V5.0.0 is deprecated (see also https://github.com/natemcmaster/CommandLineUtils/issues/533)

BREAKING CHANGE: commandline option names are updated and deprecated options are dropped.

option name --output is used instead of former --out. This avoids conflict with out parameter modifier and is compliant to recommendations [Options as parameters](https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#options-as-parameters)

image


Bertk:
@coderpatros
Hi,  I guess **GitGuardian** bot created false-positives because many of the configurations item names are like „GithubToken“, „GithubBearerToken“ and „BaseUrlUserPassword“. Looks like the tool is just looking for some patterns but a name of a variable is not a secret :confused:

Bertk:
@coderpatros Hi, how can I resolve this annoying GitGuardian false-positve of a historical commit?

PR #721, opened 2023-06-09
use System.CommandLine instead McMaster.Extensions.CommandLineUtils

Bertk:
solves #750

BREAKING CHANGE: commandline option names are updated and deprecated options are dropped.

option name --output is used instead of former --out. This avoids conflict with out parameter modifier and is compliant to recommendations [Options as parameters](https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#options-as-parameters)

image


Bertk:
The deprecated flag was removed from **System.ComponentModel.Annotations** V5.0.0
image

Bertk:
@coderpatros Sorry, but I do not know how to handle the GitGuardian false positiv.

Bertk:
I cannot find a secret in this code :wink: . Maybe GitGuardian checks GitHub API calls ...

https://github.com/Bertk/cyclonedx-dotnet/commit/d6580b1edc6800031b621571d3f8ec4def6ea50e#diff-81d67f852eeccaa952c271eb87893044L58

CodeTigerCloud:
Hi @Bertk,
is there a reason why you left this on Draft?
I'd like to use this PR as a base for the next release.

We should be able to handle the false positives with an ignore-list like here:
https://github.com/GitGuardian/ggshield/blob/main/.gitguardian.yaml

Bertk:
> Hi @Bertk, is there a reason why you left this on Draft? I'd like to use this PR as a base for the next release.
> 
> We should be able to handle the false positives with an ignore-list like here: https://github.com/GitGuardian/ggshield/blob/main/.gitguardian.yaml

Hi @CodeTigerCloud ,
I rebased the PR, fixed the code coverage actions and removed draft status.
The _GitGuardian Security Checks_ are annoying and I do not waste time with false positives of this tool. Thanks for the` .gitguardian.yaml` hint but I think a GitGuardian account is required to get the correct hash and I have no plans to touch this tool again.

CodeTigerCloud (approved):

PR #726, opened 2023-06-21
Fix framework resolution when resolving dependencies

### LINKED ISSUE #725: KeyNotFoundException with 2.8.0
After updating to 2.8.0, I can not use CycloneDX anymore... This has not happened with 2.7.0.
```
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.Text.Json.JsonElement.GetProperty(String propertyName)
   at CycloneDX.Services.ProjectAssetsFileService.SetIsDevDependency(String packageName, String targetRuntime, JsonElement jsonContent) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/ProjectAssetsFileService.cs:line 113
   at CycloneDX.Services.ProjectAssetsFileService.GetNugetPackages(String projectFilePath, String projectAssetsFilePath, Boolean isTestProject, Boolean excludeDev) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/ProjectAssetsFileService.cs:line 63
   at CycloneDX.Services.ProjectFileService.GetProjectNugetPackagesAsync(String projectFilePath, String baseIntermediateOutputPath, Boolean excludeTestProjects, Boolean excludeDev, String framework, String runtime) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/ProjectFileService.cs:line 145
   at CycloneDX.Program.OnExecuteAsync(CommandLineApplication app) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Program.cs:line 255
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.InvokeAsync(MethodInfo method, Object instance, Object[] arguments)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.OnExecute(ConventionContext context, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.Conventions.ExecuteMethodConvention.<>c__DisplayClass0_0.<<Apply>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
   at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync[TApp](CommandLineContext context, CancellationToken cancellationToken)
   at CycloneDX.Program.Main(String[] args) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Program.cs:line 140
   at CycloneDX.Program.<Main>(String[] args)
```

Falco20019:
Fixes #725 

I also adjusted the tests to reflect the possibility of having runtimes being set to avoid regression in the future.

### DIFF
--- a/CycloneDX/Program.cs
+++ b/CycloneDX/Program.cs
@@ -28,7 +28,6 @@ using CycloneDX.Services;
 using System.Reflection;
 using System.Linq;
 using CycloneDX.Interfaces;
-using System.Text.Json;
 
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("CycloneDX.Tests")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("CycloneDX.IntegrationTests")]
@@ -128,9 +127,8 @@ namespace CycloneDX
 #endregion options
 
         internal static IFileSystem fileSystem = new FileSystem();
-        internal static readonly IJsonDocs jsonDoc = new JsonDocs();
         internal static readonly IDotnetCommandService dotnetCommandService = new DotnetCommandService();
-        internal static readonly IProjectAssetsFileService projectAssetsFileService = new ProjectAssetsFileService(fileSystem, dotnetCommandService, () => new AssetFileReader(), jsonDoc);
+        internal static readonly IProjectAssetsFileService projectAssetsFileService = new ProjectAssetsFileService(fileSystem, dotnetCommandService, () => new AssetFileReader());
         internal static readonly IDotnetUtilsService dotnetUtilsService = new DotnetUtilsService(fileSystem, dotnetCommandService);
         internal static readonly IPackagesFileService packagesFileService = new PackagesFileService(fileSystem);
         internal static readonly IProjectFileService projectFileService = new ProjectFileService(fileSystem, dotnetUtilsService, packagesFileService, projectAssetsFileService);
--- a/CycloneDX/Services/ProjectAssetsFileService.cs
+++ b/CycloneDX/Services/ProjectAssetsFileService.cs
@@ -23,6 +23,7 @@ using System.Linq;
 using CycloneDX.Interfaces;
 using NuGet.Versioning;
 using System.Text.Json;
+using NuGet.Frameworks;
 
 namespace CycloneDX.Services
 {
@@ -60,6 +61,7 @@ namespace CycloneDX.Services
                     var runtimePackages = new HashSet<NugetPackage>();
                     foreach (var library in targetRuntime.Libraries.Where(lib => lib.Type != "project"))
                     {
+                        var targetAlias = TargetFrameworkToAlias(targetRuntime.TargetFramework);
                         var package = new NugetPackage
                         {
                             Name = library.Name,
@@ -67,8 +69,8 @@ namespace CycloneDX.Services
                             Scope = Component.ComponentScope.Required,
                             Dependencies = new Dictionary<string, string>(),
                             // get value from project.assets.json file ( x."project"."frameworks".<framework>."dependencies".<library.Name>."suppressParent") 
-                            IsDevDependency = SetIsDevDependency(library.Name, targetRuntime.Name, frameworksProperties),
-                            IsDirectReference = SetIsDirectReference(library.Name, targetRuntime.Name, frameworksProperties)
+                            IsDevDependency = SetIsDevDependency(library.Name, targetAlias, frameworksProperties),
+                            IsDirectReference = SetIsDirectReference(library.Name, targetAlias, frameworksProperties)
                         };
 
                         // is this a test project dependency or only a development dependency
@@ -95,9 +97,8 @@ namespace CycloneDX.Services
 
             return packages;
         }
-        public bool SetIsDirectReference(string packageName, string targetRuntime, JsonElement jsonContent)
+        public bool SetIsDirectReference(string packageName, string framework, JsonElement jsonContent)
         {
-            string framework = TargetFrameworkToAlias(targetRuntime);
             JsonElement packageProperties;
             if (jsonContent.GetProperty(framework).GetProperty("dependencies").TryGetProperty(packageName, out packageProperties))
             {
@@ -106,9 +107,8 @@ namespace CycloneDX.Services
             }
             return false;
         }
-        public bool SetIsDevDependency(string packageName, string targetRuntime, JsonElement jsonContent)
+        public bool SetIsDevDependency(string packageName, string framework, JsonElement jsonContent)
         {
-            string framework = TargetFrameworkToAlias(targetRuntime);
             JsonElement packageProperties;
             if (jsonContent.GetProperty(framework).GetProperty("dependencies").TryGetProperty(packageName, out packageProperties))
             {
@@ -127,8 +127,9 @@ namespace CycloneDX.Services
         ///     netcoreapp3.1  => netcoreapp3.1
         ///     net6.0  => net6.0
         /// </summary>
-        private string TargetFrameworkToAlias(string target)
+        private string TargetFrameworkToAlias(NuGetFramework framework)
         {
+            var target = framework.ToString();
             if (!string.IsNullOrEmpty(target))
             {
                 target = target.ToLowerInvariant().TrimStart('.');
--- a/CycloneDX/Services/ProjectAssetsFileService.cs
+++ b/CycloneDX/Services/ProjectAssetsFileService.cs
@@ -22,8 +22,8 @@ using CycloneDX.Models;
 using System.Linq;
 using CycloneDX.Interfaces;
 using NuGet.Versioning;
-using System.Text.Json;
-using NuGet.Frameworks;
+using NuGet.LibraryModel;
+using NuGet.ProjectModel;
 
 namespace CycloneDX.Services
 {
@@ -32,14 +32,12 @@ namespace CycloneDX.Services
         private readonly IFileSystem _fileSystem;
         private readonly IDotnetCommandService _dotnetCommandService;
         private readonly Func<IAssetFileReader> _assetFileReaderFactory;
-        private readonly IJsonDocs _assetJsonObject;
 
-        public ProjectAssetsFileService(IFileSystem fileSystem, IDotnetCommandService dotnetCommandService, Func<IAssetFileReader> assetFileReaderFactory, IJsonDocs assetJsonObject)
+        public ProjectAssetsFileService(IFileSystem fileSystem, IDotnetCommandService dotnetCommandService, Func<IAssetFileReader> assetFileReaderFactory)
         {
             _fileSystem = fileSystem;
             _dotnetCommandService = dotnetCommandService;
             _assetFileReaderFactory = assetFileReaderFactory;
-            _assetJsonObject = assetJsonObject;
         }
 
         public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string projectAssetsFilePath, bool isTestProject, bool excludeDev)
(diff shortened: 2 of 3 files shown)


CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -60,15 +61,16 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
                     {
+                        var targetAlias = TargetFrameworkToAlias(targetRuntime.TargetFramework);
                         var package = new NugetPackage
                         {
                             Name = library.Name,
                             Version = library.Version.ToNormalizedString(),
                             Scope = Component.ComponentScope.Required,
                             Dependencies = new Dictionary<string, string>(),
                             // get value from project.assets.json file ( x."project"."frameworks".<framework>."dependencies".<library.Name>."suppressParent") 
-                            IsDevDependency = SetIsDevDependency(library.Name, targetRuntime.Name, frameworksProperties),
```
Falco20019, line 70: `targetRuntime.TargetFramework` is adding the runtime identifier, which is not expected.

BlythMeister:
LGTM :)

NextFreeUsername (changes requested):
The major change made here is moving the TargetFrameworkToAlias to a higher place to only run once instead of twice in each "SetIsXxx"-method, but that code actually does work.
The problem lies within the project.assets.json not having a "dependencies"-element beneath project > frameworks > "someframework".
I believe the change therefore should be to also use TryGetProperty here or test for its existence in another way.

Falco20019:
No, the main change was to not use .Name which would include the RuntimeIdentifier. So TryParse alone would not fix it. I only additionally moved it out to avoid code duplication and improve performance. If that's not wanted, I can also adjust the PR to have it as minimal as possible.

CycloneDX/Services/ProjectAssetsFileService.cs
Falco20019: @NextFreeUsername This is the main change. `targetRuntime.TargetFramework.ToString()` instead of `targetRuntime.Name`. The name is internally constructed by combining the `TargetFramework` with `RuntimeIdentifier`.

NextFreeUsername:
Of course avoiding code duplication and performance improvement is always a good idea, no comments there.
But I don't think this alone fixes the issue? Resolving the framework starting from "targets" isn't the problem in my case, it is the lack of "dependencies" beneath "frameworks".

image

Falco20019:
Ok, in your case it's another issue then that didn't appear in my scenario :) But I will gladly adjust the PR on Monday with tests and a fix for that case if not done by any of the authors until then. Thanks for clarification, I thought you misread the change since it wasn't too obvious due to the refactoring being in the same commit.

andreycha:
The fix doesn't cover following use case:

```json
{
  "version": 3,
  "targets": {
    ".NETFramework,Version=v4.8": {
    }
  },
  "libraries": {
  },
  "projectFileDependencyGroups": {
    ".NETFramework,Version=v4.8": [
    ]
  },
  "packageFolders": {
  },
  "project": {
    "version": "0.0.0.1",
    "restore": {
    },
    "frameworks": {
      "net48": {
      }
    },
    "runtimes": {
    }
  }
}
```

Falco20019:
> 

It would be better to use an existing implementation like https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs but this is nothing that I will fix with my PR since it's out of scope. Please create a separate issue (I'm not a maintainer here).

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -60,15 +61,16 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
                     var runtimePackages = new HashSet<NugetPackage>();
                     foreach (var library in targetRuntime.Libraries.Where(lib => lib.Type != "project"))
                     {
+                        var targetAlias = TargetFrameworkToAlias(targetRuntime.TargetFramework);
```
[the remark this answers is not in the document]
Falco20019: Addressed in the new commit.

NextFreeUsername (approved):

Falco20019:
> The fix doesn't cover following use case:
> 
> ```json
> {
>   "version": 3,
>   "targets": {
>     ".NETFramework,Version=v4.8": {
>     }
>   },
>   "libraries": {
>   },
>   "projectFileDependencyGroups": {
>     ".NETFramework,Version=v4.8": [
>     ]
>   },
>   "packageFolders": {
>   },
>   "project": {
>     "version": "0.0.0.1",
>     "restore": {
>     },
>     "frameworks": {
>       "net48": {
>       }
>     },
>     "runtimes": {
>     }
>   }
> }
> ```

I gave it another look, since it just felt badly implemented with all the code duplication and manual parsing of the JSON while the LockFile is used for other things. I therefore reimplemented the feature and removed some code added in 2.8.0 not necessary anymore. By this change, `net48` should also be supported now.

CycloneDX/Program.cs
```diff
@@ -128,9 +127,8 @@ class Program {
 #endregion options
 
         internal static IFileSystem fileSystem = new FileSystem();
-        internal static readonly IJsonDocs jsonDoc = new JsonDocs();
         internal static readonly IDotnetCommandService dotnetCommandService = new DotnetCommandService();
-        internal static readonly IProjectAssetsFileService projectAssetsFileService = new ProjectAssetsFileService(fileSystem, dotnetCommandService, () => new AssetFileReader(), jsonDoc);
```
Falco20019, line 133: This is reverted to the 2.7.0 state, as this change is not needed by my approach.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -48,27 +47,25 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
             if (_fileSystem.File.Exists(projectAssetsFilePath))
             {
                 var assetFileReader = _assetFileReaderFactory();
-                string jsonContent = assetFileReader.ReadAllText(projectAssetsFilePath);
-                JsonDocument assetFileObject = _assetJsonObject.Parse(jsonContent);
-                // get all direct nuget dependencies of the project
-                JsonElement frameworksProperties = assetFileObject.RootElement.GetProperty("project").GetProperty("frameworks");
-
                 var assetsFile = assetFileReader.Read(projectAssetsFilePath);
```
Falco20019, line 50: Since the `assetFile` is already parsed as `LockFile` from the data hat `assetJsonObject` would usually be, I replaced this by getting the framework dependencies from the lock file.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -48,27 +47,25 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
 
                 foreach (var targetRuntime in assetsFile.Targets)
                 {
                     var runtimePackages = new HashSet<NugetPackage>();
+                    var targetFramework = assetsFile.PackageSpec.GetTargetFramework(targetRuntime.TargetFramework);
```
Falco20019, line 55: Before the PR, `targetRuntime.Name` was used to resolve the framework. This sadly includes the `RuntimeIdentifier` which is not part of the package spec and would lead to exceptions (my initial report).

The `LockFile` already has support for resolving this without the need of manually trying to map this, making `TargetFrameworkToAlias` obsolete.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -48,27 +47,25 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
+                    var dependencies = targetFramework.Dependencies;
+
                     foreach (var library in targetRuntime.Libraries.Where(lib => lib.Type != "project"))
                     {
+                        var libs = dependencies.FirstOrDefault(ld => ld.Name.Equals(library.Name));
```
Falco20019, line 60: `FirstOrDefault` will return `null` if the library is not part of the `dependencies` section.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -95,51 +92,13 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
         {
-            string framework = TargetFrameworkToAlias(targetRuntime);
-            JsonElement packageProperties;
-            if (jsonContent.GetProperty(framework).GetProperty("dependencies").TryGetProperty(packageName, out packageProperties))
-            {
-                // every direct reference has target property
-                return packageProperties.TryGetProperty("target", out _);
-            }
-            return false;
+            return dependency?.ReferenceType == LibraryDependencyReferenceType.Direct;
```
Falco20019, line 97: I made sure by checking the implementation that this is indeed how the target is mapped to. It also has the benefit that it is already mapped onto exactly what we expect (and catches some other side-cases not caught by the existing code).

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -95,51 +92,13 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
-                target = target.ToLowerInvariant().TrimStart('.');
-                var targetParts = target.Split(",version=v");
-                if (targetParts.Length == 2)
-                {
-                    return string.Join("", targetParts);
-                }
-                return targetParts[0];
-            }
-            return null;
+            return dependency != null && dependency.SuppressParent != LibraryIncludeFlagUtils.DefaultSuppressParent;
```
Falco20019, line 101: The existance of `suppressParent` is not a good metric for checking if it's a development dependency, as it CAN be set in other occasions as well. The existing code would have failed if `PrivateAssets="Build;Analyzers"` was set on the `PackageReference` (keeping `ContentFiles`).

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -95,51 +92,13 @@ public HashSet<NugetPackage> GetNugetPackages(string projectFilePath, string pro
...
-        /// <summary>
-        /// Converts an asset file's target framework value into a csproj target framework value.
-        ///
-        /// Examples:
-        ///     .NetStandard,Version=V3.1 => netstandard3.1
-        ///     .NetCoreApp,Version=V3.1  => netcoreapp3.1
-        ///     netcoreapp3.1  => netcoreapp3.1
-        ///     net6.0  => net6.0
-        /// </summary>
-        private string TargetFrameworkToAlias(string target)
```
Falco20019, line 130: As @NextFreeUsername mentioned, this lacks support for .NET Framework (and others).

Falco20019 (commented):
Some write-up about the intentions behind the changes I made.

CycloneDX/Services/ProjectAssetsFileService.cs
Falco20019: Technically, the ones interesting here would be `Compile`, `Runtime` and `Native`. If any of these would NOT be suppressed, at least some files from the package WOULD end up in the deployable. `DefaultSuppressParent` contains  `Build;ContentFiles;Analyzers` are marked private assets (see https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets and https://github.com/NuGet/NuGet.Client/blob/6.7.0.110/src/NuGet.Core/NuGet.LibraryModel/LibraryIncludeFlagUtils.cs#L16-L20 for reference).

NextFreeUsername (approved):
Very neat found I must say to use those NuGet-libraries instead of own logic.
This code looks a lot better than it used to.
Thanks a lot!

BlythMeister:
@coderpatros LGTM?

Falco20019:
@coderpatros Could you please have a look? The release is sadly broken like this and we had to blacklist the version in our Artifactory to avoid users getting it and breaking our environment.

coderpatros (approved):

jonasaebersold:
Nice! We had the same issue and the new release (2.8.1) fixed it. Thank you all for fixing and reviewing the issue.

NextFreeUsername:
Thanks a lot @Falco20019! Works like a charm on the new 2.8.1.

Issue #588, opened 2022-09-28
.NET6 Client-Side Libraries Missing from BOM


curnsey:
I noticed today that a .NET6 project in Dependency Track was missing Bootstrap and jQuery components.  Project is a basic .NET6 webapp generated from CLI with default folder structure.

wwwroot
--css
--js
--lib
----boostrap
------dist
----jquery
------dist

I generated a libman.json file and updated the bootstrap and jquery references for the client side libraries thinking maybe that would get picked up and processed.

```{
  "version": "1.0",
  "defaultProvider": "cdnjs",
  "libraries": [
    {
      "provider": "cdnjs",
      "library": "bootstrap@5.2.1",
      "destination": "wwwroot/lib/bootstrap/dist"
    },
    {
      "provider": "cdnjs",
      "library": "jquery@3.6.1",
      "destination": "wwwroot/lib/jquery/dist"
    }
  ]
}
```

Attempted to recreate the BOM after the references were restored and the same result - missing client-side libraries. Any thoughts, suggestions, advice? TIA

ArjenKorevaar:
Any thoughts or update on this ?

ArjenKorevaar:
I wouldn't mind spending some time trying to add the client-side libraries from libman.json files to the BOM, but would like to know if this idea is supported by the maintainers first.

Bertk:
Cyclone **dotnet** tool uses immediate build output (obj folder) to identify nuget packages and transitive dependencies.

Please check [cyclonedx-npm tool](https://www.npmjs.com/package/@cyclonedx/cyclonedx-npm).

ArjenKorevaar:
> Cyclone **dotnet** tool uses immediate build output (obj folder) to identify nuget packages and transitive dependencies.

CycloneDX **dotnet** tool should create SBoMs from **dotnet** projects. Imho that should include Libman as well.
 
> Please check [cyclonedx-npm tool](https://www.npmjs.com/package/@cyclonedx/cyclonedx-npm).

Afaik the CycloneDX tool for NPM does not scan Libman packages.

Bertk:
Sorry, I did not recognize the **libman** detail and we use npm for angular client. 

I guess there is no [CycloneDX support](https://cyclonedx.org/tool-center/) for [libman tool](https://devblogs.microsoft.com/dotnet/library-manager-client-side-content-manager-for-web-apps/#what-libman-is-not).

[...]

PR #732, opened 2023-07-14
Added support for Libman

ArjenKorevaar:
Added support for client-side libraries using Libman

Also see #588

### DIFF
--- a/CycloneDX.Tests/CdnjsServiceTests.cs
+++ b/CycloneDX.Tests/CdnjsServiceTests.cs
@@ -1,0 +1,83 @@
+// This file is part of CycloneDX Tool for .NET
+//
+// Licensed under the Apache License, Version 2.0 (the “License”);
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an “AS IS” BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) OWASP Foundation. All Rights Reserved.
+
+using System.Net;
+using System.Threading.Tasks;
+using CycloneDX.Models;
+using CycloneDX.Services;
+using RichardSzalay.MockHttp;
+using Xunit;
+
+namespace CycloneDX.Tests
+{
+    public class CdnjsServiceTests
+    {
+        [Fact]
+        public async Task GetComponentAsync_ReturnsComponent()
+        {
+            // Arrange
+            var mockResponseContent = @"{
+                ""name"": ""testpackage"",
+                ""description"": ""Test Description"",
+                ""author"": ""Test Author"",
+                ""license"": ""MIT"",
+                ""homepage"": ""https://www.google.com""
+            }";
+
+            using var mockHttp = new MockHttpMessageHandler();
+            mockHttp.When($"{CdnjsService.BaseUrl}/libraries/testpackage")
+                            .Respond("application/json", mockResponseContent);
+
+            var cdnjsService = new CdnjsService(mockHttp.ToHttpClient());
+            var package = new LibmanPackage(LibmanProvider.cdnjs) { Name = "testpackage", Version = "1.0.2" };
+
+            // Act
+            var component = await cdnjsService.GetComponentAsync(package).ConfigureAwait(false);
+
+            // Assert
+            var packageUrl = Utils.GeneratePackageUrl(PackageType.Libman, "testpackage", "1.0.2");
+
+            Assert.Equal(Component.Classification.Library, component.Type);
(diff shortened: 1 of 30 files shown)


CodeTigerCloud:
I just wanted to thank @ArjenKorevaar for putting effort into this essential feature.  Hope it's being released soon.

CodeTigerCloud:
I tested your changes in our Infrastructure and found a small problem. 
We upload our SBOMs to Dependency Track, which compares the components with vulnerability databases by using PURL or CPEs. 

The PURL has a type-segment, which is filled with "libman" eg: pkg:libman/datatables.net-fixedheader-dt@3.2.0"

However, our DependencyTrack did not manage to match those PURLs to vulnerabilities. Since those often state a different Type. 
I learned that the type generally describes the ecosystem the package comes from. The ASP.NET Library Manager only downlaods those packages from such ecosystems, but doesn't run any repository online. The `defaultProvider ` and `provider `in the libman.json tell Libman where to get the package from. In our case, that's usually CDNJS.

I was told, that CDNJS basically only copies packages from NPM and removes Node.js functionality. So I suggest that NPM should be used as a type. I don't know though if there are other providers (YARN?) that could also be relevant. Maybe ultimately a mapping for providers to ecosystems is required.

[PURL Spec](https://github.com/package-url/purl-spec)
[PURL-Types supported by Sona-Types seems like a collection of good types to use](https://help.sonatype.com/iqserver/product-information/referencing-package-url-%28purl%29-and-component-identifiers)

The qualifier of the PURL could be used to store the provider, but it seems to be a good practice to keep the PURLs as short as possible. I think there are argument for storing the provider in a bom as well as against it.

I'm sorry that this might become more complicated.

ArjenKorevaar:
Thank you for testing!

Libman supports 4 different package sources: CDNJS, Unpkg, JSDelivr and Filesystem.
Filesystem is ignored by this code, since it just copies files from a given source directory on the local filesystem that could contain anything.

Most packages on CDNJS seem to be sourced from NPM, and if they do CDNJS will tag them as such and provide a link to NPM. But some packages are not. Take for instance JQuery Inview (I don't know this package, I just noticed on CDNJS that it doesn't have NPM as source attribute)

https://www.npmjs.com/package/jquery-inview
https://cdnjs.com/libraries/jquery.inview

These seem to be completely different packages (different versions, authors, Github sources) and also shows that not all packages from CDNJS are from NPM.

Unpkg states they are a "content delivery network for everything on NPM" so I guess thats clear.

From what I have read about JSDelivr, they too serve packages from NPM, but also straight from Github.

I don't mind it getting more complicated, but I'm not sure what to do here. I understand why things like SCA tools or vulnerability scanners have problems with an unknown package source `libman`, but I don't think the correct solution would be to label them as NPM packages.

CodeTigerCloud:
Thank you for the clarifications. I am not quite fond of the JS-package world.

If CDNJS et at also deliver other packages that are not from NPM, I totally agree that it would not be correct to just set the type to NPM. However, I think the type then should be CDNJS/Unpkg/JSDelivr instead of libman, as libman - to my understanding - is basically a client to download from those services. Ignoring Filesystem sounds reasonable to me.

Also, entering NPM gives me plenty of node.js specific vulnerabilities that are only relevant for backends - not optimal.

It appears to me, that a proper SCA infrastructure for JS-frontend packages doesn't exist. I mean, it makes sense, since the harm that broken frontend components can cause is limited compared to backend components.

I think the CycloneDX/OWASP-teams need to figure some things out for themselves first.

ArjenKorevaar:
I could try to at least tag the packages that are known to originate from NPM as such.

As I mentioned, Unpkg packages should always come from NPM.
CDNJS offers metadata that could be used to determine the origin.
For JSDelivr the package url might be of use.

But that leaves the question open as to what to do with packages that are not from NPM.

CodeTigerCloud:
I don't see any satisfying way to solve this right now. Whatever we settle as the format of the PURL can later turn out to be the wrong decision. 

So I suggest, we don't decide on a PURL-format for libman components yet. Meaning, we leave the PURL out, but instead write the source of the package as externalReference.
Later, when DependencyTrack or the SCA-field in general caught up on how to handle frontend-packages, we can decide then make a decision without having to revert functionality.

Does that sound fair?

ArjenKorevaar:
We have decided to move away from using Libman, since we're running into issues with support on all levels. Even if you manage to produce an SBoM with Libman packages, chances are that any (SCA) tool that you plan to use it with won't accept them or won't be able to provide you with any vulnerability, license or other information. Making it kinda useless and even a potential security risk.

mtsfoni:
Hey Arjen,

thanks for informing me. I am curious, what other way did you chose for your client side libraries that also has sort of SCA support? We face the same problem in my day job, where I think I still run a version of your PR-branch, but I just get an insane amount of node.js related false positives.

I'd still like the opportunity to get the libman libraries optionally added to the sbom, so I hope you don't mind me still using this code for a future release.

ArjenKorevaar:
Hi @mtsfoni, we're moving to NPM for all client side libraries. This does create a new problem where we need to run multiple types of CycloneDX tools and merge the SBoM's to get the complete picture. But we can bundle that logic in standard components and scripts that will make our life a little easier.

Feel free to use whatever you like from the code!

Issue #669, opened 2023-03-29
Test projects are not excluded with option "exclude-test-projects"


AlbertoMonteiro:
Opening a new issue because the previous one was closed, but didnt fixed all the cases as I explained in my comment: 
https://github.com/CycloneDX/cyclonedx-dotnet/issues/363#issuecomment-1489385428

Bertk:
Please add the property `<IsTestProject>true</IsTestProject>`  in the test csproj file. As far as I know cyclonedx-dotnet does not merge all props and target files.

schoofseggl:
I can confirm this bug. In Directory.Build.Props we set the property `IsTestProject` to all test projects: 

```
	<PropertyGroup Condition="$(MSBuildProjectName.EndsWith('.Tests')) or $(MSBuildProjectName.EndsWith('.IntegrationTests'))">
		<IsPackable>False</IsPackable>
		<IsTestProject>True</IsTestProject>
	</PropertyGroup>
```

But during the run it was analyzed even though we called it with the option to ignore test projects. When we set it directly in the csproj file, it is skipped as expected.

Bertk:
The **cyclonedx-dotnet** tool uses solution, project and intermediate output file `project.assets.json`. The tool is executed **after a build is executed** and does not use or create any derived project property settings from build target.

Therefore the definition of the _*IsTestProject*_ property in a *.props or *.target file which is resolved during **build target** execution **cannot** be used for the _exclude-test-projects_ option.

This limitation for **exclude-test-projects** option should be documented in README.md.

>"cyclonedx-dotnet tool option **exclude-test-projects** resolves IsTestProject property value from project files after a build is executed. Therefore `<IsTestProject>true</IsTestProject>` shall be set in csproj file or in an explicit import property file within the csproj file without conditions.

schoofseggl:
This is technically understood - but from a user perspecitve I am now forced to apply an error prone + manual process (to add the flag everywhere and don't forget that in the future) where there is already a better and central place. 

What about having an --ignore-pattern commandline paramter, where I can specify pattern to ignore projects that should not be scanned?

Bertk:
A new feature with a command line option `filter projects` could the list of options is already pretty huge. 

Anyway, #736 might be a better solution which also allows to eliminate some low level XML document handling.

t4m45:
[...]

PR #733, opened 2023-07-18
add exclude-test-projects limitation

Bertk:
#669
- disable redundant remark-lint check for README.md
- `.editorconfig` defines markdown format (indent_size)


Bertk:
solved with #736

PR #748, opened 2023-08-16
Nuget package from private feed not found on nuget.org results in error ("Central Directory corrupt.")

dreammaster38:
Hi,

I just stumbled about an error while trying to create SBOM from nuget packages not found on public nuget.org.
The error message was "Central Directory corrupt." because the package couldn't be downloaded from nuget.org by default.
Now i just check if the package really exists on nuget.org before trying to read any info.

Feel free to integrate this bugfix into your release if you like.

Kind regards
Thomas


CodeTigerCloud:
Hello dreammaster38,

unfortunately, you didn't sign-off your commits, thus I cannot merge them. Please consider signing them off.

See here for some more information: https://github.com/CycloneDX/cyclonedx-dotnet/pull/748/checks?check_run_id=15953566656

PR #752, opened 2023-08-17
Generate BOM metadata timestamp

### LINKED ISSUE #751: Feature request: Generate timestamp or set timestamp with command line parameter
Most of the metadata that can be imported via the `-imp` command line parameter is static, except of the version number and the timestamp, which may differ from build to build. 

The version number can already be set independently with the `-sv` command line parameter. It would be good to also either always generate a timestamp for the BOM (as it is part of the NTIA minimum requirements) or to set a timestamp via an additional command line parameter, so that all information held in the metadata file can be kept static.

I don't see any reason why the timestamp should not simply be generated by CycloneDX for every build. A command line parameter may be more flexible but possibly superfluous. Maybe both options could be supported, i.e. if no value is given for a command line parameter `set-timestamp`, then a timestamp is generated.

logicaloud:
Fixes #751: Automatically generate a BOM metadata timestamp if none is provided with the metadata.

This approach does not require any additional command line parameters.

### DIFF
--- a/CycloneDX/Program.cs
+++ b/CycloneDX/Program.cs
@@ -416,6 +416,12 @@ namespace CycloneDX
                 bom.Metadata.Component.BomRef = $"{bom.Metadata.Component.Name}@{bom.Metadata.Component.Version}";
             }
 
+            // Automatically generate a timestamp if none is provided with the metadata
+            if (bom.Metadata.Timestamp == null)
+            {
+                bom.Metadata.Timestamp = DateTime.UtcNow;
+            }
+
             AddMetadataTool(bom);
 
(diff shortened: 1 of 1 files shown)


logicaloud:
Sorry, I have missed the signature. Probably not worth worrying about as only a few lines of code would need copying. Please feel free to close without merging.

CodeTigerCloud:
Hi logicaloud,

any chance you simply add the sign-off? Then I can easily take this into the next release.

See here: https://github.com/CycloneDX/cyclonedx-dotnet/pull/752/checks?check_run_id=15995619348

logicaloud:
Thanks for considering the pull request. Adding the sign-off was less complicated than I thought :) I think all should be good to go now.

CodeTigerCloud (approved):

Issue #750, opened 2023-08-17
Replace "Microsoft.DotNet.PlatformAbstractions" with existing .NET platform implementation


Bertk:
"Microsoft.DotNet.PlatformAbstractions" nuget package can be removed after a one line update.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/platformabstractions-package-removed

This line in [CycloneDX.Tests/ProgramTests.cs](https://github.com/CycloneDX/cyclonedx-dotnet/blob/47967d7e7eccb21e0b6779fb076376a5db5b7b9c/CycloneDX.Tests/ProgramTests.cs#L104C1-L104C114) 

```csharp 
string resourcePath = Path.Join(ApplicationEnvironment.ApplicationBasePath, "Resources", "metadata");
```

with

```csharp 
string resourcePath = Path.Join(AppContext.BaseDirectory, "Resources", "metadata");
```

PR #753, opened 2023-08-18
eliminate removed package #750

Bertk:
#750


CodeTigerCloud (commented):
Hi Bertk,
can you quickly tell me why it also removes coverlet.msbuild? Is it not being used?

Bertk:
@CodeTigerCloud Sorry for the late reply. Code coverage analysis is not available in CI build or PR check and coverlet.collector is recommend because coverlet.msbuild has some known issues. Test projects should not use both coverlet packages.

test job needs this:

```
dotnet test --collect:"XPlat Code Coverage"

or with coverlet.msbuild

dotnet test /p:CollectCoverage=true
```

I also noticed some errors in CI build which are related to missing mandatory CycloneDX argument.
see https://github.com/CycloneDX/cyclonedx-dotnet/actions/runs/5898990612/job/16000868797?pr=753

Hint: _a generate and publish code coverage sample_
https://github.com/Bertk/arcade-light-sample/blob/main/.github/workflows/dotnet.yml

PR #762, opened 2023-09-22
Replace licenseID NOASSERTION with null

### LINKED ISSUE #525: NOASSERTION as LicenseId in generated SBOMs
When generating SBOM from .csproj files it seems that when the license can be parsed correctly ( its ID), `NOASSERTION` value is added instead, this causes an error when trying to consume the bom (validating) using the cyclonedx-java-core library, it says that that ID is not a valid one which is true since that value is not in the list the library uses.

In this case, maybe the way to proceed would be to ignore that license or not generated the bom?

CodeTigerCloud:
There are cases where the GitHub licence call returns a JSON that has a spdxId and a license name.

Example: https://api.github.com/repos/aspnet/AspNetWebStack/license
`  "license": {
    "key": "other",
    "name": "Other",
    "spdx_id": "NOASSERTION",
    "url": null,
    "node_id": "MDc6TGljZW5zZTA="
  }`
  
This is problematic for two reason:

- A cycloneDX licence tag must have either a name or an spdx_id, not both.
- NOASSERTION is not a valid value for the spdx_id enumeration.

Obviously, NOASSERTION is not a meaningful value, which makes an easy solution to just replace it with null. In the given example, the licence is more or less correctly identified by its name "other" as it cannot be matched to a licence.

For the future, it might be worth the consideration that the class GithubService should only return licence objects that adhere to the standard.

Fixes #525

### DIFF
--- a/CycloneDX.Tests/GithubServiceTests.cs
+++ b/CycloneDX.Tests/GithubServiceTests.cs
@@ -493,5 +493,30 @@ namespace CycloneDX.Tests
             Assert.Equal("Test License", license.Name);
             Assert.Equal("https://licenceurl.com/", license.Url);
         }
+
+        [Fact]
+        public async Task GitLicense_ReplacesNoAssertionWithNull()
+        {
+            //See also https://github.com/CycloneDX/cyclonedx-dotnet/issues/525
+
+            var mockResponseContent = @"{
+                ""license"": {                    
+                    ""name"": ""Other"",
+                    ""spdx_id"": ""NOASSERTION""                  
+                }, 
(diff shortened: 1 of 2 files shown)


stephanruhland-netzebw (commented):
Hi,
an empty id field also results in an invalid BOM. if the node id would be removed, the file would be valid (according to https://cyclonedx.github.io/cyclonedx-web-tool/validate 1.4)
Kind regards

stephanruhland-netzebw:
Thank you for your response. 
I just wanted to bring it up, we are eagerly awaiting for the fix. :)

PR #789, opened 2023-12-26
Update args

mtsfoni:


mtsfoni:
new:

-ipr ---include-project-references
-ed
-fn
-sr

deprecated:
-d
-r
-f

i think bert also removed some deprecated old ones, that must also go into the changelog

Issue #786, opened 2023-12-17
After updating to CycloneDX.Core 6.0: Add CycloneDX as Component instead of as Tool(deprecated) to the Metadata.


CodeTigerCloud:
In function AddMetadataTool CycloneDX is added as a tool inside the metadata-section, the tool node is flagged as deprecated in CycloneDX 1.5 though.

The Core-library 6.0 is not yet being used.

mtsfoni:
#804

PR #804, opened 2023-12-28
Tools/tool is deprecated in cdx standard. Using tools/components instead.

mtsfoni:


mtsfoni:
#786

firetech:
Is there any reason why this is left in limbo? The fact that cyclonedx-dotnet output is using the deprecated syntax causes it to overwrite any other tools data when merging with newer formats using [cyclonedx-cli](https://github.com/CycloneDX/cyclonedx-cli).

PR #825, opened 2024-01-09
Removed a code snippet in ProjectFileService.cs that caused a crash and has no use

ag-lls:
We encountered a problem with the current 3.0.4 version of the tool. We had a weird crash of the tool in some of our build pipelines. It seems to be a problem regarding duplicate keys in a dictionary. 

We debugged the code and found a foreach loop that added dependencies into a unused object property. We removed the code since we don't see the use in it and removing the code also fixed our crash.

This is the stack trace of the error we encountered:
```
Unhandled exception: System.ArgumentException: An item with the same key has already been added. Key: Microsoft.IO.RecyclableMemoryStream
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at CycloneDX.Services.ProjectFileService.RecursivelyGetProjectDotnetDependencysAsync(String projectFilePath, String baseIntermediateOutputPath, Boolean excludeTestProjects, String framework, String runtime) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/ProjectFileService.cs:line 226
```

### DIFF
--- a/CycloneDX/Services/ProjectFileService.cs
+++ b/CycloneDX/Services/ProjectFileService.cs
@@ -221,11 +221,6 @@ namespace CycloneDX.Services
             {
                 var projectDotnetDependencys = await GetProjectDotnetDependencysAsync(project.Path, baseIntermediateOutputPath, excludeTestProjects, framework, runtime).ConfigureAwait(false);
 
-                foreach (var dependency in projectDotnetDependencys)
-                {
-                    project.Dependencies.Add(dependency.Name, dependency.Version);
-                }
-
                 DotnetDependencys.UnionWith(projectDotnetDependencys);
             }
 


ag-lls:
If this PR is accepted, what would be a possible timeframe for a 3.0.5 fix? Asking since this crash is a bit blocking to us 😸

mtsfoni:
I think the code you remove is needed to build the dependency graph. The unused object is not unused but merged into the return value.

Unfortunately you didn't provide a lot of information, I don't completely understand how this can happen. Could you provide some more information?

Apparently, the `GetProjectDotnetDependencysAsync` in line 222 returns at least one dependency double, which shouldn't be the case. If you help me understand why that happens, I should be able to fix the problem.

ag-lls:
You are right, the code has use, we made a mistake. Obviously the `projectReferences` object gets used after the loop. Sorry for that - happen when it's late 😄  For the actual issue (the crash) i will open a proper issue. Thanks for the quick response.

Issue #833, opened 2024-01-18
"Unable to locate valid bom ref" when AssemblyName is != project name in referenced project


ag-lls:
We are facing an issue where the the cyclonedx-dotnet app exits with the message `Unable to locate valid bom ref` and code 7.

The problem can be triggered by the following setup:
- Project1 references Project2
- Project2 has the `AssemblyName` Test.Project2
- Project2 references Project3

@mtsfoni I have once again created a test project for you to reproduce the problem.
https://github.com/ag-lls/RefTest

The problem does not trigger if Project2 does not reference Project3 or if Project2's `AssemblyName` is equal to the project name exactly. 

We think the issue maybe has something to do with how a referenced project is getting read out as a project reference and as a dependency.
<br/>
As a project reference, the name of the project is used (in this case Project2), while as a dependency the `AssemblyName` is used (in this case Test.Project2). This results in the two entries not getting caught by the `UnionWith` in line 254 in ProjectFileService.cs.

![image](https://github.com/CycloneDX/cyclonedx-dotnet/assets/140407621/81be50c2-64ce-4c99-af3b-23c7bf28ab48)

![image](https://github.com/CycloneDX/cyclonedx-dotnet/assets/140407621/62c1a91c-f75f-4a6b-b3bd-6c5fe07fc5e1)

This seems to be the case because for a dependency (`getDotnetDependencys`), the tool reads out the projects `project.assets.json` file, in which the "projectName" field has the AssemblyName as a value.

mtsfoni:
What argument do you call it with?

I stumbled over the same problem when using -rs and -ipr in a solution that uses packages.config after updating from 3.0.4 to 3.0.5 and already built a test-case.

> We think the issue maybe has something to do with RecursivelyGetProjectReferencesAsync(?).
And in some of the lists/dicts used in RecursivelyGetProjectDotnetDependencysAsync, sometimes projects appears two times, once "normal" with a normal key/name and a version and once with a not "normal" key/name (the full project path) and a missing version. We think that this is the root of the problem, since for that "broken" entry no bom-ref can be found.

#834 should be a fix for this. 
But I want to test some more cases before making a version as I feel that by fixing one hole I am always opening another one.

ag-lls:
@mtsfoni i just updated the issue with a better description.

mtsfoni:
Should be the same issue indeed. 
Interesting aspect about the AssemblyName, but it should also be utilized with this fix.

[...]

PR #834, opened 2024-01-18
Fix: #833 Error with includePackageReference when ProjectReference-Graph is at least 2 levels deep.

mtsfoni:


mtsfoni:
This basically need some caching for the results from reading the csproj, as currently it reads the same project-file multiple times.

However, it might be reasonable to change this directly to Buildalyzer.

mtsfoni:
If project twos assembly name is different from its project name, it will appear twice in the Components and the dependency graph is also wrong

Bertk:
> However, it might be reasonable to change this directly to Buildalyzer.

Buildalyzer supports also some caching:

> You can add all projects in a solution to the AnalyzerManager by passing the solution path as the first argument of the AnalyzerManager constructor. This will parse the solution file and execute GetProject() for each of the projects that it finds.
>
>Calling GetProject() again for the same project path will return the existing ProjectAnalyzer. You can iterate all the existing project analyzers with the IReadOnlyDictionary<string, ProjectAnalyzer> property AnalyzerManager.Projects.

PR #844, opened 2024-01-28
843 scope property is always being set to required

mtsfoni:


thompson-tomo:
@mtsfoni will this fix also result in dependencies of the dev dependency also having their scope classified as excluded as I am needing that fixed as part of this.

mtsfoni:
You are right, it doesn't. That also seems a little trickier as the asset file, doesn't give the information directly. 

Looks like I'd have to check the dependency graph and can mark one of those a dev dependency if every path towards the root include at least one dev dependency in-between.

Here D would be a dev dependency:
```
  B (dev) - D
 /
A
```


Here D cannot be a dev dependency:
```
  B (dev) - D
 /
A -  C  - D
```

Here, however, D would be a dev dependency again (all paths to A have a dev dependency on the way): 
```
  B (dev) - D
 /
A -  C (dev) - D
```


@thompson-tomo do you happen to know a package in the gallery that will automatically be added as development dependency and has dependencies by itself? Most of those package have no dependencies by themselves.

thompson-tomo:
I agree, it isn't as easy as would be hoped. So the main dev dependency package i am looking for a solution for would be **Microsoft.VisualStudio.Web.CodeGeneration.Design** & **Microsoft.EntityFrameworkCore.Tools** i am fairely sure atleast one of them is automatically added as a dev dependency,

mtsfoni:
I will certainly solve this, as I consider it a bug. But I need to shift some things around in the process to make it work, so this is bigger task.

mtsfoni:
> @mtsfoni will this fix also result in dependencies of the dev dependency also having their scope classified as excluded as I am needing that fixed as part of this.

Question: Do you need them specifically to be `scope.excluded` or is it also fine, if they just don't appear in the sbom at all? I learned that `scope.excluded` actually is meant for components that are necessary to use the software, but are not included in the scope of delivery.

thompson-tomo:
> Question: Do you need them specifically to be `scope.excluded` or is it also fine, if they just don't appear in the sbom at all? I learned that `scope.excluded` actually is meant for components that are necessary to use the software but are not included in the scope of delivery.

I don't specificially need the scope to be set to excluded and my comment was in response to the changes you had pushed. Happy for them to simply not appear in the SBOM especially if it makes things easier to fix.

thompson-tomo:
> Question: Do you need them specifically to be `scope.excluded` or is it also fine, if they just don't appear in the sbom at all? I learned that `scope.excluded` actually is meant for components that are necessary to use the software, but are not included in the scope of delivery.

I don't specificially need the scope set to excluded and my comment was in response to the changes you had pushed. Happy for them to not appear in the SBOM

Issue #842, opened 2024-01-28
Set additional properties when generating sbom for nugets


thompson-tomo:
When generating a SBOM the below properties should be set to provide a richer experience & better information.
- **Group** the first segment of the package name
- **supplier** the url of the repository used to source package. Not sure if possible
- **publisher** same as author

mtsfoni:
> Group the first segment of the package name

This bases on the assumption that packages are named by a certain convention/best practice that is not enforced. I don't think that this applies to all packages.

> publisher is the owner as specified in Nuget, if empty use author

Sounds not unreasonable. Interestingly though, on the NuGet website you usually see an owner e.g. [here](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore), but in the .nuspec-file I didn't see a filled owner node yet.
I would really love using the one from the .nuspec-file but I am little hesitant having to parse the NuGet website for that information. Which did you mean with "as specified in Nuget"?

thompson-tomo:
@mtsfoni i have just pushed a draft PR #846 which shows the publisher being set. 

In relation to the Group you are correct it is not enforced but recently nuget has started using verified prefixes to help improve security by providing visual indicator that packages are coming from a reputable source and who they claim to be.

PR #846, opened 2024-01-29
#842 Add additional properties to SBOM

thompson-tomo:
This request now sets:

- the publisher in the sbom based on the owner and if it is null it will use the author

Closes:

- #842


thompson-tomo:
@mtsfoni as per the nuspec reference on microsoft (https://learn.microsoft.com/en-us/nuget/reference/nuspec) owners is deprecated and instead should use authors moving forward.

Issue #847, opened 2024-02-06
Exclude developer dependencies is not reflecting real runtime dependencies


gotztibor:
# The NuGet package assets are not considered correctly when developer dependencies are excluded.
The following assets are considered to have "runtime" dependency: 
* LibraryIncludeFlags.Runtime,
* LibraryIncludeFlags.Native,
* LibraryIncludeFlags.ContentFiles

Upon generating the dependencies without the developer dependencies (only availabe at build, but not at runtime) the dependencies are not calculated correctly.

Example project file:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <OutputType>exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <!-- runtime true -->
    <PackageReference Include="Google.Protobuf" Version="3.25.2" PrivateAssets="analyzers;build" />
    <!-- runtime true with transitive dependencies -->
    <PackageReference Include="log4net" Version="2.0.15" ExcludeAssets="none" />
    <!-- runtime false, without transitive dependencies -->
    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" ExcludeAssets="all" />
    <!-- runtime true -->
    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" IncludeAssets="runtime" ExcludeAssets="native;contentFiles;build" />
    <!-- runtime true -->
    <PackageReference Include="Serilog" Version="3.1.1" PrivateAssets="all" IncludeAssets="runtime" />
  </ItemGroup>

</Project>
```

Expected outcome:
As the Microsoft.Extensions.Logging is excluded fully, this will be just a developer dependency, and also it's transitive dependencies must not be shown as well. Also the build output folder and the default publish folder and the possibly generated NuGets package  are not containing those NuGets and not referencing them.

gotztibor:
Please see [Pull Request 848](https://github.com/CycloneDX/cyclonedx-dotnet/pull/848)

pinkfloydx33:
I'm pretty sure this is the same issue; I just noticed that the following gets dropped by `--exclude-dev`:

```
<PackageReference Include="Microsoft.EntityFrameworkCore" Version="10.0.0" PrivateAssets="none" />
```

But... it shouldn't as it's is a full dependency. The `PrivateAssets=none` is required so that analyzers flow to consumers of a library from transitive dependencies

PR #848, opened 2024-02-06
#Issue 847 - Fix runtime dependencies if developer dependencies are excluded

gotztibor:
Please see [#Issue847](https://github.com/CycloneDX/cyclonedx-dotnet/issues/847)


gotztibor:
Dear Michael, finally I applied the proposal You made, and I am now marking only dependencies to be dev dependencies instead of forcibly removing them. Actually I think, my solution also covers the problem of the mentioned case [here](https://github.com/CycloneDX/cyclonedx-dotnet/pull/844#issuecomment-1913933677).

The serious case, because of I started to jump on the code and check how developer dependencies are handled was, that in case of we have more fine-graned control on the package assets with (ExcludeAssets,PrivateAssets,IncludeAssets)
I experienced the following:
- some of the references libraries were falling out (into dev dependencies), which were actually hard runtime dependencies (available in the bin, publish folder) eg. PrivateAssets="analyzers;build",
- some of them were additionally falling in, which were really not runtime dependencies at all

Please would You be so kind to check again the current solution proposal?
The way now is working:
-  end of the dependency resolution it is recursively traversing through the package dependencies (starting from the direct one's as roots) collecting the "real-hard-runtime" dependencies which will be present in the bin,publish folders
- all of the dependencies excluding the above will be marked as developer dependencies
- with this solution also the direct and indirect dependencies (even if the same package) is handled

Cases:
- D is a developer dependency, as that is achievable only throughout a dev dependency
```
  B (dev) - D
 /
A
```

- D is not a dev dependency as that will be part of the hard-runtime dependencies through the A -C -D path
```
  B (dev) - D
 /
A -  C  - D
```

- D is a dev dependency, because there is no path from root to it without visiting a dev dependency
```
  B (dev) - D
 /
A -  C (dev) - D
```
Even in the case above the dependency B would be present as runtime dependency as it could be achieved according to the dependency graph throughout a path without visiting a (dev) Node.

Really-really thanks a lot in advance!

mtsfoni:
At the first glance it looks good. I will test the code in the weekend.

I will have a few nitpicky change request regarding style.

Also, I'll need you to sign-off your contribution for [legal reason](https://developercertificate.org/).
[Here it describes](https://github.com/CycloneDX/cyclonedx-dotnet/pull/848/checks?check_run_id=21298026619) pretty good how you can do it retrospectively for your commits. However, you can just do this when everything else is done.

I will try to get your code released quickly. Should be in the next 2 weeks, hopefully.

Thanks for the contribution, especially as it took a little work off my back (there is still a lot to do for this tool).

gotztibor:
Dear Michael,

   sorry for my limited knowledge, but please could You share with me the proposals about the style, because this case I try to commit those changes/adaptations as well.

Really-really thanks a lot for Your efforts in advance!
Thanks for helping me out!

mtsfoni:
Hey Götz,
there is one aspect that isn't fully clear to me. We have the private asset:
```    
<!-- runtime true -->
<PackageReference Include="Serilog" Version="3.1.1" PrivateAssets="all" IncludeAssets="runtime" />    
```

Private asset means to my understanding, please correct me if I am wrong, it belongs to the output of this project, but other project referring this project (via package- or projectReference) will not output this private asset.

So by that understanding, generally your comment `<!-- runtime true -->` would apply. However, the code currently marks serilog as development dependency. 

But there is an extra twist: What if the SBOM is generated for a NuGet-(library)-package? Will the private asset be part of the package or not? Because if it isn't, marking it as development dependencies would be correct again. In that cycloneDX might need a way to just mark it as a private dependency and then figure out later (maybe by the cli-arguments) if it is a dev dependency or not.

By the way: forget about the nitpicks, as I already checked the branch out, I will just do them by myself later. It's just about preference. For example, [here ](https://github.com/CycloneDX/cyclonedx-dotnet/pull/848/files#diff-f6b4594a17fe40dd6acf8e163eaf8ec1286dfba6fb0c5bc8955763b10d29c890R111)I would just prepare a return value for more clarity instead of changing the contents of the list inside the function.

PR #878, opened 2024-05-23
[Issue] XML Serialization of Bom with multiple licenses produces invalid XML

nicolaihenriksen:
This really should be logged as an issue rather than a PR, but doing a PR allows me to extend an existing validation test-case to provoke the error that I am seeing.

The `CsvHelper` nuget package is listed on nuget.org with this license info:
image

This can also be confirmed by looking in the *.nuspec file of the package:
image

The BOM (all versions I have dealt with so far) schema definition states that the `<licenses>` element should appear 0 or 1 times. This is why the serialized Bom is invalid, because the output looks like the snippet below where there are 2 `licenses` elements instead of nesting both `license` elements into a single "collection":

```xml
﻿<?xml version="1.0" encoding="utf-8"?>
<bom xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" ... version="1" xmlns="http://cyclonedx.org/schema/bom/1.5">
  <metadata>
    ...
  </metadata>
  <components>
    <component type="library" bom-ref="pkg:nuget/CsvHelper@32.0.3">
      ...
      <licenses>
        <license>
          <id>MS-PL</id>
        </license>
      </licenses>
      <licenses>
        <license>
          <id>Apache-2.0</id>
        </license>
      </licenses>
      ...
    </component>
    ...
  </components>
  ...
</bom>
```

Expected something like this instead:
```xml
<licenses>
  <license>
    <id>MS-PL</id>
  </license>
  <license>
    <id>Apache-2.0</id>
  </license>
</licenses>
```

I suspect the issue may even be in the [cyclonedx-dotnet-library](https://github.com/CycloneDX/cyclonedx-dotnet-library) rather than in this repository, but it was easier to demonstrate the issue here.

When I run the test suite including my change below, you can see that it passes the JSON variant, but fails the XML. I suspect the same will happen in the pipeline upon opening this PR. I have ticked "Allow edits by maintainers" so feel free to add the fix directly on my fork/branch if you think that is appropriate (and even possible).
image

### DIFF
--- a/CycloneDX.Tests/ValidationTests.cs
+++ b/CycloneDX.Tests/ValidationTests.cs
@@ -34,6 +34,7 @@ namespace CycloneDX.Tests
                 new() { Name = "LibGit2Sharp", Version = "0.27.0-preview-0096" },
                 new() { Name = "NLog", Version = "4.7.7" },
                 new() { Name = "RestSharp", Version = "106.11.7" },
+                new() { Name = "CsvHelper", Version = "32.0.3" },
             };
 
             var mockProjectFileService = new Mock<IProjectFileService>();


mtsfoni:
most likely related to:
https://github.com/CycloneDX/cyclonedx-dotnet-library/pull/218

The generator tools creates a List of licenses in the Component and fills multiple licenses in that.
Looks alright so far.

nicolaihenriksen:
@mtsfoni You are absolutely correct, that seems to be the same issue. I must not have been thorough enough when searching for existing issues; perhaps because it seems there are mostly PRs opened, not really issues.

I will close this as a duplicate of the PRs below (or covered by):
https://github.com/CycloneDX/cyclonedx-dotnet-library/pull/218
https://github.com/CycloneDX/cyclonedx-dotnet-library/pull/187

PR #902, opened 2024-08-06
Fix Error with Uppercase Characters in Version Strings

JohnHunhoff:
Resolve an issue where versions with uppercase characters (e.g., 1.0.0-Beta, 1.0.0.BETA.1) caused CycloneDX to throw errors. The
problem occurred because the NuGet cache converted version strings to lowercase for path names. This caused the .nuspec file to not be found.

### DIFF
--- a/CycloneDX/Services/NugetV3Service.cs
+++ b/CycloneDX/Services/NugetV3Service.cs
@@ -80,11 +80,12 @@ namespace CycloneDX.Services
             if (string.IsNullOrEmpty(name) || string.IsNullOrEmpty(version)) { return null; }
 
             var lowerName = name.ToLowerInvariant();
+            var lowerVersion = version.ToLowerInvariant();
             string nuspecFilename = null;
 
             foreach (var packageCachePath in _packageCachePaths)
             {
-                var currentDirectory = _fileSystem.Path.Combine(packageCachePath, lowerName, NormalizeVersion(version));
+                var currentDirectory = _fileSystem.Path.Combine(packageCachePath, lowerName, NormalizeVersion(lowerVersion));
                 var currentFilename = _fileSystem.Path.Combine(currentDirectory, lowerName + _nuspecExtension);
                 if (_fileSystem.File.Exists(currentFilename))
                 {


JohnHunhoff:
To provide additional information, the error message is "Central Directory corrupt," and it originates from the file NugetV3Service.cs. This error occurs in all versions greater than 2.3. To test, you just need a project that depends on a NuGet package with an uppercase character in the version..

mtsfoni:
Do you happen to have an example at hand of a public package that has an uppercase character in its version?

ecaisse:
This PR fixes https://github.com/CycloneDX/cyclonedx-dotnet/issues/603, at least when it comes to people having issues with Linux-based build agents.

> Do you happen to have an example at hand of a public package that has an uppercase character in its version?

I did not find any public NuGet packages that uses an uppercase in it's version. It may be that the public feed automatically sets all versions to lowercase.

However, for private feeds, those are possible. I have such a case using Azure Artifact feeds, where we use a timestamp as part of the version number for internal prerelease packages, similar to this format: `1.0.0-20240918T123456`. Note the capital `T` as part of the timestamp. However, in the NuGet global cache, the folder name has a lowercase `T`.

In a Windows environment, that's not a problem since paths are case-insensitive. When `NugetV3Service.GetCachedNuspecFilename` check if the file exists, it returns true despite casing being different. Since it's in the global cache, `nuspecFilename` is not null and `NugetV3Service.GetNuspec` reads the `.nupkg` file from disk as expected.

That is not the case with Linux. In that case, the file cannot be found because of the casing difference, and the check fails. In `NugetV3Service.GetNuspec`, `nuspecFilename` is null, which calls `resource.CopyNupkgToStreamAsync`. This call fails silently since the return value is not handled (I can ran some tests and this returns false, meaning the operation failed). The stream is not a valid package (most likely empty), which then throws an error when trying to parse the Nupkg archive.

To the best of my knowledge, the NuGet client normalizes versions to lowercase in the global cache when restoring packages. The [GlobalPackagesFolderUtility](
https://github.com/NuGet/NuGet.Client/blob/e082545cbba30ec8dc9b759f2d9beea8aabc6215/src/NuGet.Core/NuGet.Protocol/Utility/GlobalPackagesFolderUtility.cs#L34) class instantiates a [VersionFolderPathResolver](https://github.com/NuGet/NuGet.Client/blob/e082545cbba30ec8dc9b759f2d9beea8aabc6215/src/NuGet.Core/NuGet.Packaging/VersionFolderPathResolver.cs#L29) with the parameter `isLowerCase` equal to `true` (using the default parameter value). The [Normalize](https://github.com/NuGet/NuGet.Client/blob/e082545cbba30ec8dc9b759f2d9beea8aabc6215/src/NuGet.Core/NuGet.Packaging/VersionFolderPathResolver.cs#L189) method for the NuGet version will return a lowercase version to be used as part of the path on disk.

JohnHunhoff:
I don't have an example, the error occurred to me with private packages self hosted on directory or nexus repository manager in linux. The only public package I know of is https://www.nuget.org/packages/Unity/5.9.0-RC1.

yorschor:
Imagine my amusement when i find your pull request after finding the exact same fix (character for character) after 2 days of debugging 😂 
@Bjornej exact same situation here ^^ 

Same question. Is there anything that need to be done before this can get merged? :)

mtsfoni (approved):

mtsfoni:
Very busy these days I'll try to make a new version with this in the weekend

PR #903, opened 2024-08-08
Prevent `Unable to locate valid bom ref for package` error when using a range and there are multiple versions of a package referenced

tl-maisie-sadler:
## Context

I have come across an issue while using this tool to generate an SBOM for one of our larger solutions.

When the following occurs:
1. Package dependency is specified with range syntax, e.g. `[1.4.0, 1.4.0]`
2. Multiple versions of the package referenced throughout the solution

We get an error message as below, even though there are packages available.

`Unable to locate valid bom ref for TrueLayer.Observability.Abstractions [1.4.0, 1.4.0]`

## Solution

The proposed solution chooses the first referenced package, rather than assuming none exists if the found packages != 1.

### DIFF
--- a/CycloneDX.Tests/ProgramTests.cs
+++ b/CycloneDX.Tests/ProgramTests.cs
@@ -18,12 +18,17 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Abstractions;
 using System.IO.Abstractions.TestingHelpers;
+using System.Runtime.Serialization.Formatters;
+using System.Security.Cryptography;
 using System.Threading.Tasks;
 using CycloneDX.Interfaces;
 using CycloneDX.Models;
+using CycloneDX.Services;
 using Moq;
 using Xunit;
+using static CycloneDX.Models.Component;
 using XFS = System.IO.Abstractions.TestingHelpers.MockUnixSupport;
 
 namespace CycloneDX.Tests
@@ -75,7 +80,7 @@ namespace CycloneDX.Tests
                 .Setup(s => s.GetSolutionDotnetDependencys(It.IsAny<string>(), It.IsAny<string>(), It.IsAny<bool>(), It.IsAny<string>(), It.IsAny<string>()))
                 .ReturnsAsync(new HashSet<DotnetDependency>());
 
-            Runner runner = new Runner(fileSystem: mockFileSystem, null, null, null, null, null, solutionFileService: mockSolutionFileService.Object, null);            
+            Runner runner = new Runner(fileSystem: mockFileSystem, null, null, null, null, null, solutionFileService: mockSolutionFileService.Object, null);
 
             RunOptions runOptions = new RunOptions
             {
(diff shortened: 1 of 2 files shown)


mtsfoni:
The core issue is, that a c#-solution is not one software-solution (as in tool, library or service) but rather an arbitrary collection of one or more software-solution or parts of such.

The issue comes from the idea, that a software-solution can only have one version of a dependency at one time. Which is basically correct for C#. 

This bases on the assumption that you would create a SBOM (in the context of this tool) for a single software-solution. Which makes sense.

Running this tool against a c#-solution, now assumes that multiple software-solution are actually just one because they are arbitrarily grouped, might find multiple dependencies with same name but different version - which basically is correct, because they are used in different software-solutions. 

If one just discards all versions but the first, one creates an incomplete SBOM. In my opinion it's actually a flaw that you can run this tool against a .sln-file and I'd recommend running it against the root-component (.csproj) of each included software-solution separately and potentially - if necessary - merge them after. If the project uses PackageReferences inside the project file, transitive dependencies will be detected and resolved to the correct version.

The correct solution for this problem, would be, to simply add all versions of that library to the sbom, because all of them might be in use by some of the parts of the software solution. We can't easily know when creating a sbom for solutions.

mtsfoni:
I am unable to reproduce the underlying problem. 

Could you describe, how I need to setup a solution to run into the error?

aledalmas:
Install a package with a dependency with version 1.1.1 inside a project (In example Newtonsoft).
Install another package with the same packet as dependency with version 1.1.2 inside a project.

VS will tell you to install one of those 2 version directly inside your csproj to solve the issue.

In my case I have a package with Microsoft.Extensions.DependencyInjection.Abstractions = 8.0.1 and another package with dependency Microsoft.Extensions.DependencyInjection.Abstractions 8.0.2.

VS told me to install the 8.0.2 directly inside my project and now I'm stuck with this error when running the tool:

Found 384 packages
Unable to locate valid bom ref for Microsoft.Extensions.DependencyInjection.Abstractions [8.0.1, 8.0.1]
subprocess exited with status 7
subprocess exited with status 7
Error: building at STEP "RUN dotnet CycloneDX ./XXX.sln -o . etc,": exit status 7

ilehtoranta:
Any update to this? I am running into the same issue.

Found 399 packages
Unable to locate valid bom ref for Microsoft.CodeAnalysis.Workspaces.Common [5.0.0, 5.0.0]

Would be nice if there were a flag to ignore errors and generate a BOM anyway.

mtsfoni:
@ilehtoranta could you provide a minimal version of the csproj and the cyclone-dx call for reproduction? That is what is blocking me here

ilehtoranta:
@mtsfoni here is the minimal version from the project I am working on.

[CDXTest.zip](https://github.com/user-attachments/files/26566563/CDXTest.zip)

**Run**
dotnet cyclonedx CDX.sln

**Failure output**
Found 80 packages
Unable to locate valid bom ref for Microsoft.CodeAnalysis.Workspaces.Common [5.0.0, 5.0.0]

mtsfoni:
Awesome, I will look into it shortly

mtsfoni:
This has been fixed in [v6.1.1](https://github.com/CycloneDX/cyclonedx-dotnet/releases/tag/v6.1.1). It would be great if you could give it a try and let us know if it resolves the issue for you.

ilehtoranta:
Yes, it works now.

Found 399 packages

Creating CycloneDX BOM
Writing to: [...]bom.xml

Great that you got it nailed. Thank you :)

PR #907, opened 2024-09-18
Update CycloneDX to 8.0.0 and BOM validation to v1_6

MarioAllegro:
Updated FunctionalTestHelper.cs and ValidationTests.cs to support BOM validation for Specification Version v1_6 in both JSON and XML formats. Upgraded CycloneDX.Core package version in Directory.Packages.props from 7.0.1 to 8.0.0. Also corrected a minor formatting issue in Directory.Packages.props.


disshishkov:
There is command to generate bom file with previous spec version 1.5?

Issue #911, opened 2024-10-24
Incorrect dependency structure on different target framework


jesperolsson-se:
A dependency may reference different dependency versions based on the project's target framework. When scanning a solution where two projects with different target frameworks reference the same dependency, the generated SBOM may incorrectly label a transitive dependency as a direct dependency.

**Steps to reproduce**

1. Given a fresh solution
2. Add a net6.0 project (P1)
3. Let P1 reference `Microsoft.Data.SqlClient@5.2.1`
4. Add a net8.0 project (P2)
5. Let P2 reference `Microsoft.Data.SqlClient@5.2.1`
6. Generate an SBOM for the solution

**Result**

- Microsoft.Data.SqlClient@5.2.1
  - System.Runtime.Caching@6.0.0
- System.Runtime.Caching@8.0.0

PR #912, opened 2024-10-25
chore: test for #911

jesperolsson-se:


jesperolsson-se:
@mtsfoni 
Now that the bug is confirmed, would you mind merging this?
Since the test is ignored it shouldn't affect your CI, and this way it won't become stale in case of refactoring.

PR #777, opened 2023-11-17
#774 Support for .NET 8

Matiszak:

PR #909, opened 2024-10-17
Enable `RollForward` to run on the latest .NET SDKs

nil4:
Enables installing and running `CycloneDX` on runtimes/SDKs newer than those that it targets when published. (For instance, allows v4.0.0 to run with only .NET SDK 9.0 installed, and no 8.0, 7.0 or 6.0 present.)

ref. https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#major-version-runtime-roll-forward
ref. https://github.com/dotnet/sdk/issues/26824

Examples of other tools which have this setting enabled:
- `dotnet-format` (MS): https://github.com/dotnet/format/blob/ad8125a6fc036fe1eb4e57aa604a79f2d98d5aa0/src/dotnet-format.csproj#L11  -- added in https://github.com/dotnet/format/commit/7a5df81d357992e66bf88b6dd664d70569bd7f21
- `dotnet-ef` (MS): https://github.com/dotnet/efcore/blob/486047c7422ec04530ca66c7c55256fa89b0ffc2/src/dotnet-ef/dotnet-ef.csproj#L24 -- added in https://github.com/dotnet/efcore/commit/11e7fc66f23edf3df7eb41d3b8ed311cb2dd2b60
- `docfx` (MS): https://github.com/dotnet/docfx/blob/dbd57048f68411fa39aeea9ccc6c00dc88ecde9f/src/docfx/docfx.csproj#L5 -- added in https://github.com/dotnet/docfx/commit/c641c038a9ee20a25500c454c57339cd98ebf8cd
- my own `dotnet-xdt`: https://github.com/nil4/dotnet-transform-xdt/blob/6272bb1152e8535ec40da5b0d68ff950bc13735b/dotnet-xdt/dotnet-xdt.csproj#L47 -- added in https://github.com/nil4/dotnet-transform-xdt/issues/42

PR #914, opened 2024-11-15
feat: add .NET 9 Support

nathan-mittelette:
Add support for .NET 9 based on this PR : https://github.com/CycloneDX/cyclonedx-dotnet/pull/777


nil4:
Adding a new `net-9.0` target to this tool would work, but I would argue that doing so would be inefficient and ultimately, unnecessary. 

The tool currently targets .NET 6, 7, and 8. Adding a new .NET target would be inefficient because it would increase the package size by ~33% without providing new APIs or features. 

It is not necessary because there is a relatively obscure, but straightforward, way for end-users to install and use the tool as-is with .NET 9.

Consider the scenario where only .NET 9 SDK is installed (and `[...]` denotes elided output to keep this brief):

```cmd
> dotnet --version
9.0.100

> dotnet tool install --global CycloneDX
You can invoke the tool using the following command: dotnet-CycloneDX
Tool 'cyclonedx' (version '4.0.0') was successfully installed.

> dotnet cyclonedx --version
You must install or update .NET to run this application.

App: C:\Users\nil4\.dotnet\tools\dotnet-CycloneDX.exe
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '8.0.0' (x64)
.NET location: C:\Program Files\dotnet\

The following frameworks were found:
  9.0.0 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]

[...]
To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=8.0.0&arch=x64&rid=win-x64&os=win10
```

ref. https://youtu.be/4iEqqPZKDC0?t=2081 end-users can explicitly add `--allow-roll-forward` when installing the tool to work around this issue:

```cmd
> dotnet tool uninstall --global CycloneDX
Tool 'cyclonedx' (version '4.0.0') was successfully uninstalled.

> dotnet tool install --global --allow-roll-forward CycloneDX
You can invoke the tool using the following command: dotnet-CycloneDX
Tool 'cyclonedx' (version '4.0.0') was successfully installed.

> dotnet cyclonedx --version
4.0.0+3d71d1f1a7609ea5c8108ca11ea07d5ad2f0222b
```

With this, `CycloneDX` can be installed and used **as-is** on .NET 9 *and future versions*.

Additionally, end-users could be spared the inconvenience of having to add `--allow-roll-forward` when installing (and contributors/maintainers rushing to add new `netX.0` targets yearly) by adding `<RollForward>Major</RollForward>` to the tool's project file.

This is what https://github.com/CycloneDX/cyclonedx-dotnet/pull/909 suggests -- that change would allow the straightforward `dotnet tool install --global CycloneDX` to work with any future .NET SDK.

maarten-kieft:
Anhy news on this? We facing issues that the docker file doesn't support .net 9. We could ofcourse create our own docker image, but we rather stay in sync.

nil4:
The `rollForward`=`major` policy was merged in https://github.com/CycloneDX/cyclonedx-dotnet/pull/909 and the subsequent release `4.1.0` includes it: https://github.com/CycloneDX/cyclonedx-dotnet/releases/tag/v4.1.0

My testing showed that installing and using version 4.1.0 worked on .NET 9 SDK. Should you face any problems, may I suggest opening a new issue to add the relevant details (e.g. stacktraces, etc.) @maarten-kieft ?

nil4:
I suppose you're referring to this line here:

https://github.com/CycloneDX/cyclonedx-dotnet/blob/3ad0483c8724b6ddd0c20506b034dde5f8555b44/Dockerfile#L1

If so, I can see how updating that base image version to a 9.0 SDK would be helpful. 

I would suggest requesting the project maintainers' feedback about this topic -- I'm not familiar with how the Docker image is created and published, nor using it myself.  My feeling is that many other changes in this PR may be too broad, if this is the only update needed.

mtsfoni:
I built a new version 4.2 with sdk:9.0 as base image - everything else should work with the `rollForward` option.

Image needs to need to have the newest version because so it can `nuget restore` the projects it is supposed to scan.

@maarten-kieft please give me feedback if this resolves your issues.

maarten-kieft:
Thanks @mtsfoni , I will try it Monday and let you know. Much appreciated!

maarten-kieft:
@mtsfoni The issue is resolved! Thank you very much!

Issue #853, opened 2024-02-14
Support providing a solution filter


KieranFoot:
It would be nice if instead of the current supported file type, we could provide solution filters to the tool. This would allow developers to hand pick the projects that they wish to be part of the BOM.

mtsfoni:
Can you describe what you understand under "solution filter".

Can you describe a case where using a project- or a solution-file is not sufficient for selecting relevant projects? What is it you are trying to achieve? Is this an A-B Problem?

KieranFoot:
It's not an issue, it's just an improvement that would assist people that are already using solution filters, but I'll proceed to describe my use case.

I currently have a monolithic solution containing code for many target architectures and operating systems. I have CI/CD pipelines for each target since the code and dependencies cannot be compiled on just any platform.

In both Visual Studio and Rider, you can create a Project Filter from a solution, this is basically a JSON file that describes both the solution and the projects to include (thus filtering the ones not included).

I would like the ability to use a solution filter as I currently use it for all relevant `dotnet` commands in place of a solution/project and it would be simpler if I could use it here too rather than having to create and manage the lists in several different places.

Also, in my opinion, the solution filter format is much simpler then a solution file to both consume and maintain manually.

For some information on filters, see [here](https://devblogs.microsoft.com/visualstudio/load-only-the-projects-you-need-with-solution-filters/)

As I say, currently, this is how I separate code to progress through my pipelines per target.

mtsfoni:
So I assume you use the cyclonedx tool on the solution-file rather than the specific project-file? 

When a solution file is targeted, the cyclonedx tool basically creates a bom for each project inside the solution and aggregates those mindlessly. This is not a very clever, smart, or even correct way to build a bom. If there are different output projects that have different versions of dependencies, it gets super messy. Yes, solution filters could reduce that problem.

However, there is an explicit way to get exactly what one needs and want:
Call the cyclonedx tool with the project file, that is the actual output (the webserver, windowsservice, .exe etc.). Unless one still uses packages.config files, cyclonedx will read all important information from the nuget-generated project.assets.json file and generate a quite accurate bom with the dependency tree.
In my workplace, we generate a seperate frontend bom and a backend bom from the respective projects inside the solution.

That being said. I am not absolute against support for solution filters, however I see calling the tool with a solution file more as a comfortable shortcut than an accurate approach of generating a bom. A program is normal one project (with dependencies); a solution is a bundle of many (usually related) projects.
[...]

PR #920, opened 2025-01-13
Add support for solution filter file (#853)

michha:
I tried different implementations and also a separate `SolutionFilterFileService.cs` but found it very clunky with sharing most of the code with `SolutionFileService.cs` so I finally came to this result.


ChristoWolf:
Hi @michha!

Any chance to get this soon?
I'd also need it :)

mtsfoni (approved):

Issue #904, opened 2024-09-02
System.IO.IOException: An attempt was made to move the position before the beginning of the stream.


SagarKodam:
Hi Team,

Command used: dotnet cyclonedx --json --out OutputPath Solutionpath --url AlternativeNuGetSourceprovided --disable-github-licenses --exclude-dev --exclude-test-projects --dotnet-command-timeout 900000

It fails with the below error - 

Analyzing: *******************.csproj
  Attempting to restore packages
  Packages restored
File not found: "***************.json", "**********************.csproj" 
  No packages found
Found 64 packages
1 packages being excluded as DevDependencies
Unhandled exception: System.IO.InvalidDataException: Central Directory corrupt.
 ---> System.IO.IOException: An attempt was made to move the position before the beginning of the stream.
   at System.IO.MemoryStream.Seek(Int64 offset, SeekOrigin loc)
   at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
   --- End of inner exception stack trace ---
   at System.IO.Compression.ZipArchive.ReadEndOfCentralDirectory()
   at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen, Encoding entryNameEncoding)
   at System.IO.Compression.ZipArchive..ctor(Stream stream, ZipArchiveMode mode, Boolean leaveOpen)
   at NuGet.Packaging.PackageArchiveReader..ctor(Stream stream, Boolean leaveStreamOpen, IFrameworkNameProvider frameworkProvider, IFrameworkCompatibilityProvider compatibilityProvider)
   at NuGet.Packaging.PackageArchiveReader..ctor(Stream stream)
   at CycloneDX.Services.NugetV3Service.GetNuspec(String name, String version, String nuspecFilename, FindPackageByIdResource resource) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 319
   at CycloneDX.Services.NugetV3Service.GetNuspec(String name, String version, String nuspecFilename, FindPackageByIdResource resource) in /home/runner/work/cyclonedx-dotnet/cyclonedx-dotnet/CycloneDX/Services/NugetV3Service.cs:line 324
[...]

PR #921, opened 2025-01-16
Added NugetConfigFileService, parsing nuget.config for package sources

GeorgRottensteiner:
Allow package lookup from a list of repositories

### DIFF
--- a/CycloneDX.Tests/FunctionalTests/FunctionalTestHelper.cs
+++ b/CycloneDX.Tests/FunctionalTests/FunctionalTestHelper.cs
@@ -39,7 +39,8 @@ namespace CycloneDX.Tests.FunctionalTests
                 It.IsAny<RunOptions>(),
                 It.IsAny<IFileSystem>(),
                 It.IsAny<IGithubService>(),
-                It.IsAny<List<string>>()))
+                It.IsAny<List<string>>(),
+                It.IsAny<HashSet<NugetInputModel>>()))
                 .Returns(nugetService);
 
             return mockNugetServiceFactory.Object;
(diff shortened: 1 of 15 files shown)


mtsfoni:
Is this related to #904? Could you give me a little more context before I review?

Also you need to sign off your commits. Here is also a detailed guide how to retrospectively: https://github.com/CycloneDX/cyclonedx-dotnet/pull/921/checks?check_run_id=35933831961

GeorgRottensteiner:
Looks like it. The issue is that NuGet.Protocol returns a non-null Stream object even if there is no file at all.
One thing that I did is add a check for that (length == 0 )
Obviously that is not enough, since there is a reason the file is not there. That's why I opted to add nuget.config parsing. Not sure if this is the proper way to record several NuGet sources. NuGet keeps shuffling options and methods unfortunately.

Also as a hint: I did not add package source authentication, albeit the models allowing for such. I did more or less the bare minimum to get it working for my use case. If that method works that might be something that ought to be added.

Will look into that signing off asap!

GeorgRottensteiner:
Not sure if I did that correctly, I have a 50% miss rate with rebasing :)
The commits look like having the intended sign-off.
Never heard of that sign-off thingy before this PR.

Need that StringBuilder issue Codacy bemoans also rectified?

GeorgRottensteiner:
I wanted to explain the other part I modified:
Since the projects I handle use more than one package source, I also replaced the single source with a list of sources. This requires gracefully handling the failure of finding a package on one source, and only abort if a package cannot be found on none of them.
In case of a package existing in several sources, first one wins. Basically the source order in nuget.config decides the priority. Which is IIRC also how NuGet handles this.

mtsfoni:
I think I understand and after skimming briefly the code looks good so far.

I'd like to set up an environment to recreate the problem and test your solution, as that is some work, it might take a little before I integrate this PR. 
But thank you for your contribution already!

PR #922, opened 2025-01-17
Log the package and version when unzipping fails

pregress:
Logging for https://github.com/CycloneDX/cyclonedx-dotnet/issues/904 to track down which package is causing it.


CycloneDX/Services/NugetV3Service.cs
```diff
@@ -316,8 +316,16 @@ private async Task<NuspecModel> GetNuspec(string name, string version, string nu
...
                     _logger, _cancellationToken);
 
-                using PackageArchiveReader packageReader = new PackageArchiveReader(packageStream);
-                nuspecModel.nuspecReader = await packageReader.GetNuspecReaderAsync(_cancellationToken);
+                try
+                {
+                    using PackageArchiveReader packageReader = new PackageArchiveReader(packageStream);
+                    nuspecModel.nuspecReader = await packageReader.GetNuspecReaderAsync(_cancellationToken);
+                }
+                catch (InvalidOperationException)
```
vaujo6y, line 324: Are you not trapping the wrong exception?

the ones in the issue and the ones i'm seeing are:
InvalidDataException

vaujo6y (commented):
``

Issue #894, opened 2024-07-10
Grpc.Tools version 2.55.1 not resolved to specific version error


Kubana:
Couple of days ago I started getting this exception when generating SBOM from one of our projects.

Dependency (Grpc.Tools) with version range ([2.55.1, )) referenced by (Name:Grpc.AspNetCore Version:2.55.0) did not resolve to a specific version.
Unable to locate valid bom ref for Grpc.Tools [2.55.1, )

What is the issue here?

mtsfoni:
Are you using the current version of cyclonedx-dotnet? I hoped we got rid of that bug.

Do you maybe have a different version of grpc.tools referenced directly?

Kubana:
I think I have but let me doublecheck

Kubana:
Currently running the dotnet tool CycloneDX version 3.0.8

mtsfoni:
With what argument do you call the tool?
Normaly a single project.assets.json should be unambiguous.
This problem can arise, if you have a solution where different projects refer to different versions of a library as the tool doesn't know which version to use.

I want to change the tool to just add all versions of that library in case, but I have no free capacity.

A workaround can be, to generate the sboms by refering the "root-component"-projects (meaning your .exe / service-backend ). I recommend doing it that way for increased accuracy anyway and handle all my projects like that. (I generate a frontend and backend bom seperately then)

Kubana:
dotnet-CycloneDX.exe {projectName}.csproj -o CycloneDXOutputs -fn {projectName}.xml -dpr

Kubana:
Problem persists even in version 4.0.0

mtsfoni:
Hey,

is this still a problem for you? I just tried but I just cannot get it reproduced

Could you describe the dependencies between the projects and exactly which version of Grpc.AspNetCore/Tools is referenced by which project?

Falco20019:
I have the same issue. I am currently looking to create a regression test for this and maybe also a fix.

Falco20019:
@mtsfoni This is exactly what the test `ProjectAssetsFileServiceTests.GetDotnetDependencys_MissingResolvedPackageVersion` tries. While the assets part is fine with keeping the dependency, `NugetV3Service.GetComponentAsync` will fail to resolve it and fail the generation. I resolved it with the linked PR to some extend that will work for my use-case and also this one. It will still fail more exotic ones like those (basically any that excludes the min value):
* (1.0,)
* (,1.0]
[...]

PR #929, opened 2025-02-18
fix: improvement on handling build-only dev dependencies

Falco20019:
Fixes https://github.com/CycloneDX/cyclonedx-dotnet/issues/894 for `--recursive` project mode and solution mode.

This also fixes the following bugs that were not reported yet to my knowledge but lead to wrong results:
* CentralPackageVersions not being considered for resolving version ranges
* `packages.config` not being considered `IsDirectReference`
* Recursive mode considering ANY dependency as direct that comes from the initial project
* Recursive mode will add old versions from references for already packages updated in main project
* Recursive mode will add dependencies of references partially as direct references

So many bugs were related to `--recursive` not working correctly.

### DIFF
--- a/CycloneDX/Runner.cs
+++ b/CycloneDX/Runner.cs
@@ -521,7 +521,10 @@ namespace CycloneDX
             foreach (var runtimePackage in runtimePackages)
             {
                 var dependencies = runtimePackage.Dependencies?.ToList();
-                if (dependencies == null) continue;
+                if (dependencies == null)
+                {
+                    continue;
+                }
 
                 foreach (var dependency in dependencies)
                 {
--- a/CycloneDX/Runner.cs
+++ b/CycloneDX/Runner.cs
@@ -27,6 +27,7 @@ using CycloneDX.Models;
 using CycloneDX.Interfaces;
 using CycloneDX.Services;
 using static CycloneDX.Models.Component;
+using NuGet.Versioning;
 
 namespace CycloneDX
 {
@@ -238,8 +239,8 @@ namespace CycloneDX
                 await Console.Out.WriteLineAsync($"{packages.Where(p => p.IsDevDependency).Count()} packages being excluded as DevDependencies");
             }
 
-
-            
+            // Resolve version ranges:
+            ResolveDependencyVersionRanges(packages);
 
             // get all the components and dependency graph from the NuGet packages
             var components = new HashSet<Component>();
@@ -511,5 +512,35 @@ namespace CycloneDX
             }
         }
 
+        /// <summary>
+        /// Updates all dependencies with version ranges to the version it was resolved to where possible.
+        /// </summary>
+        private static void ResolveDependencyVersionRanges(HashSet<DotnetDependency> runtimePackages)
+        {
+            var runtimePackagesLookup = runtimePackages.ToLookup(x => x.Name.ToLowerInvariant());
+            foreach (var runtimePackage in runtimePackages)
+            {
+                var dependencies = runtimePackage.Dependencies?.ToList();
+                if (dependencies == null) continue;
+
+                foreach (var dependency in dependencies)
+                {
+                    if (!NuGetVersion.TryParse(dependency.Value, out _) && VersionRange.TryParse(dependency.Value, out VersionRange versionRange))
+                    {
+                        var normalizedDependencyKey = dependency.Key.ToLowerInvariant();
+                        var package = runtimePackagesLookup[normalizedDependencyKey].FirstOrDefault(pkg => versionRange.Satisfies(NuGetVersion.Parse(pkg.Version)));
+                        
+                        // This can happen for development build-only dependencies.
+                        if (package == default)
+                        {
+                            Console.Error.WriteLine($"Dependency ({dependency.Key}) with version range ({dependency.Value}) referenced by (Name:{runtimePackage.Name} Version:{runtimePackage.Version}) did not resolve to a specific version. You might want to try --recursive to resolve this.");
+                            continue;
+                        }
+
+                        runtimePackage.Dependencies[dependency.Key] = package.Version;
+                    }
+                }
+            }
+        }
     }
 }
--- a/CycloneDX/Services/PackagesFileService.cs
+++ b/CycloneDX/Services/PackagesFileService.cs
@@ -22,7 +22,6 @@ using System.IO.Abstractions;
 using System.Threading.Tasks;
 using CycloneDX.Interfaces;
 using CycloneDX.Models;
-using System.Linq;
 using System;
 
 namespace CycloneDX.Services
@@ -64,6 +63,7 @@ namespace CycloneDX.Services
                                 Name = reader["id"],
                                 Version = reader["version"],
                                 IsDevDependency = reader["developmentDependency"] == "true",
+                                IsDirectReference = true,
                                 Scope = Component.ComponentScope.Required
                             };
                             await Console.Out.WriteLineAsync($"\tFound Package:{newPackage.Name}");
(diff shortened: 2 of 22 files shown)


Falco20019:
I adjusted the code to not create the dummies for now but to at least fix the broken `-rs` path that should have worked.

CycloneDX/Services/ProjectFileService.cs
```diff
@@ -240,24 +239,19 @@ public async Task<HashSet<DotnetDependency>> GetProjectDotnetDependencysAsync(st
         public async Task<HashSet<DotnetDependency>> RecursivelyGetProjectDotnetDependencysAsync(string projectFilePath, string baseIntermediateOutputPath, bool excludeTestProjects, string framework, string runtime)
         {
             var dotnetDependencys = await GetProjectDotnetDependencysAsync(projectFilePath, baseIntermediateOutputPath, excludeTestProjects, framework, runtime).ConfigureAwait(false);
-            foreach (var item in dotnetDependencys)
-            {
-                item.IsDirectReference = true;
-            }
-
```
Falco20019, line 247: This was wrong in the first place and only compensated the missing `IsDirectReference` from `PackagesFileService.GetDotnetDependencysAsync()`. So by fixing it there, this is not necessary anymore. I assume this is here from before `--includeProjectReferences` was around, but that handles this already.

CycloneDX/Services/ProjectFileService.cs
```diff
@@ -270,11 +264,18 @@ public async Task<HashSet<DotnetDependency>> RecursivelyGetProjectDotnetDependen
...
 
-                dotnetDependencys.UnionWith(projectDotnetDependencys);
+                // A higher version might already be added
+                allAddedDepedencyNames = dotnetDependencys.Select(dep => dep.Name);
+                dotnetDependencys.UnionWith(projectDotnetDependencys.Where(pr => !allAddedDepedencyNames.Contains(pr.Name)).Select(d =>
+                {
+                    // Ensure project references are not added as direct references
+                    d.IsDirectReference = false;
+                    return d;
+                }));
```
Falco20019, line 274: The current code would merge in older versions from referenced projects . In my case I got a second version of `Serilog` in which was incorrect.

`Serilog.Extensions.Hosting@8.0.0` on the referenced project used referenced `Serilog@3.1.1` but my main-project referenced already `Serilog@4.0.0` which was also set as dependency for `Serilog.Extensions.Hosting@8.0.0` in `dotnetDependencys` correctly. So `Serilog@3.1.1` was added as component but never referenced due to `Serilog.Extensions.Hosting@8.0.0` by `UnionWith` being taken from the main project.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -130,9 +131,9 @@ public bool SetIsDevDependency(LibraryDependency dependency)
         }
 
         /// <summary>
-        /// Updates all dependencies with version ranges to the version it was resolved to.
+        /// Updates all dependencies with version ranges to the version it was resolved to where possible.
         /// </summary>
-        private static void ResolveDependencyVersionRanges(HashSet<DotnetDependency> runtimePackages)
+        private static void ResolveDependencyVersionRanges(HashSet<DotnetDependency> runtimePackages, IDictionary<string, CentralPackageVersion> centralPackageVersions)
```
Falco20019, line 136: I split-up the logic to allow better resolution with `-rs` for missing dev-dependencies. This can be easily reproduced by referencing `Grpc.Tools` in any version and referencing that project by another project. The SBOM has to be generated for that second project. `Grpc.Tools` is a build-only dev dependency that is stripped by the `project.assets.json` generator as non-relevant (which cyclonedx would rely on).

So this new method tries to now only reduce the range to what was set in the central package management (if used) to be later on the `Runner.ResolveDependencyVersionRanges` doing the matching. For non-recursive mode, this will make no difference and generate the same warnings at that point (now with an added note to use recursive mode). In recursive mode, the files would have been added by that point through the package reference and therefore resolve without warnings to the correct non-ranged version.

CycloneDX/Services/ProjectAssetsFileService.cs
```diff
@@ -143,15 +144,24 @@ private static void ResolveDependencyVersionRanges(HashSet<DotnetDependency> run
...
+                            // Central package versions can already limit the expected range.
+                            if (!centralPackageVersions.TryGetValue(dependency.Key, out var cpv))
+                            {
+                                // If we don't have a CPV, ignore for now.
+                                // It will be tried to match later in the Runner again.
+                                continue;
+                            }
+
+                            // Despite the name, this could already be a non-range version.
+                            patchVersion = cpv.VersionRange.ToShortString();
```
Falco20019, line 161: `ToShortString` generated the simplest name to be used:
* 1 -> 1.0.0
* [1, ) -> 1.0.0
* (1, ) -> (1.0.0, )
* [1, 2) -> [1.0.0, 2.0.0)

Non-min-inclusive version ranges would not be resolvable by the old version of the code, but will be resolvable in the new run for recursive project or solution runs.

CycloneDX/Services/PackagesFileService.cs
```diff
@@ -64,6 +63,7 @@ public async Task<HashSet<DotnetDependency>> GetDotnetDependencysAsync(string pa
                                 Name = reader["id"],
                                 Version = reader["version"],
                                 IsDevDependency = reader["developmentDependency"] == "true",
+                                IsDirectReference = true,
```
Falco20019, line 66: Was was defaulting to `false` but packages in the config are always directly referenced by definition and would be migrated as those for `PackageReference`.

CycloneDX/Runner.cs
```diff
@@ -511,5 +512,35 @@ internal static void AddMetadataTool(Bom bom)
             }
         }
 
+        /// <summary>
+        /// Updates all dependencies with version ranges to the version it was resolved to where possible.
+        /// </summary>
+        private static void ResolveDependencyVersionRanges(HashSet<DotnetDependency> runtimePackages)
```
Falco20019, line 518: This will now resolve the packages and is basically the same code as before on the `ProjectAssetFileService`. Moving it ensures that it now really sees all packages, including build-only dev dependencies stripped by the `project.assets.json` generator.

It's important that this will still only resolve the packages for recursive project or solution mode. On non-recursive, these will still be missing and there is no way to fetch them. They COULD be added in non-recursive mode as `IsDevDependency=true` with `Scope=Exclude` but might still miss things like dependencies of the real package. So I (for now) decided against adding any workaround and only suggest the user in the warning to use `recursive` mode instead.

Falco20019 (commented):
I added some comments as it looks like a big PR. Technically I didn't change a lot besides needing to fix some more bugs in the run as they interferred with this feature.

mtsfoni:
I think I understand the problem now.

The simplest fix would be to remove a dependency if it can’t be resolved to a package. If it’s not listed in the assets file, it doesn’t exist.

SBOMs are already generated with dev-dependencies, which, while not the highest priority, are still relevant from a security standpoint. There’s also a valid reason to include recursive, transitive dev-dependencies. However, from what I understand, that would require scanning the assets files of project references recursively.

I originally saw the `-rs` option as a legacy feature meant for `packages.config`, but I can see why someone might assume it’s the right way to include dev-dependencies as described above.

Here is what I'd do now:
1. Remove dependencies that don’t exist as packages, which should resolve the current did not resolve to a specific version issues.
2. Always scan for project references (but only add them if the flag is set) and check their assets files for dev-dependencies. If dev-dependencies are excluded, they’ll be dropped. With this behaviour added, `-rs` should not be necessary for projects with `project.assets.json` any more.
3. Ignore `-rs` for projects that don’t use packages.config and display a warning.

Step two will require a bigger rewrite and will take some time, but it should also be a good opportunity to clean up some things.


I'd like to get your perspective on this, as you already dove into this problem.

Falco20019:
Should work in my case. Since I had `-ipr` set, the project would be missing the dev dependency as it wasn't in the main projects assets. If both assets are checked also without `-rs`, this sounds like a plan to me.

Also fine with your points 1 & 3

Falco20019:
Resolved by https://github.com/CycloneDX/cyclonedx-dotnet/pull/934

Issue #523, opened 2022-04-21
Add purl to component in the metadata


mr-zepol:
When generating a bom the metadata component does not have a purl, if we have the information ( coordinates), it should be able to create the purl, and even better use that as bom-ref instead of `name@version`

RodneyRichardson:
This could be helpful to generate the purl value: https://github.com/package-url/packageurl-dotnet

PR #931, opened 2025-02-24
feat: Add purl to metadata component

Falco20019:
Fixes https://github.com/CycloneDX/cyclonedx-dotnet/issues/523

I need this feature as-well. I assume it was only marked as "Not planned" due to being stale.

### DIFF
--- a/CycloneDX.Tests/FunctionalTests/RunOptionsValidation.cs
+++ b/CycloneDX.Tests/FunctionalTests/RunOptionsValidation.cs
@@ -1,0 +1,46 @@
+using System.IO;
+using System.Threading.Tasks;
+using CycloneDX.Models;
+using Xunit;
+
+
+namespace CycloneDX.Tests.FunctionalTests
+{
+    public class RunOptionsValidation
+    {
+        [Fact(Timeout = 15000)]
+        public async Task Defaults()
+        {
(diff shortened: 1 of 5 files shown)


mtsfoni:
I reopened the issue, automatically closing stale issue turned out to be a bad idea.

Without diving deeper into the topic: is it possible to solve this without the flag? Just deciding by what is inside the metadata file?

Falco20019:
I think we COULD do it always if it's the type `library` (maybe some others, definetly not for `application`). I just tried to do it non-breaking for now as a changed `bom-ref` could be considered breaking. If it was provided by `-imp` before, it will not change either way.

Falco20019:
@mtsfoni Any more thoughts?

mtsfoni:
We'll proceed with the current implementation, but for now, please keep only the long option `--set-nuget-purl` and omit the shorthand `-snp.` I have some larger changes planned and prefer to hold off on committing to new shorthands options at this stage.

The idea of automatically applying it to libraries is interesting, and we should revisit it later when introducing other breaking changes.

Could you remove the shorthand? Once that's done, I'll merge and release this change. 
Thanks!

Falco20019:
@mtsfoni Thanks :)

mtsfoni (approved):

PR #933, opened 2025-03-07
feat: Enable .slnx solutions to be analyzed

MMonrad:
This Pull Request aim to support the new solution format .slnx, that was released together with .NET 9.0.200.

It goes with the old approach and uses regex to read the projects path, before start reading the solution file it checks if it is a .sln or .slnx and then call the macthing private method.

### DIFF
--- a/CycloneDX.Tests/SolutionFileServiceTests.cs
+++ b/CycloneDX.Tests/SolutionFileServiceTests.cs
@@ -31,6 +31,28 @@ namespace CycloneDX.Tests
     public class SolutionFileServiceTests
     {
         [Fact]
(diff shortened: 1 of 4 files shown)


mtsfoni:
Hi, could you [sign off ](https://github.com/CycloneDX/cyclonedx-dotnet/pull/933/checks?check_run_id=38362264686) your commits, then I can merge and release your changes.

MMonrad:
> Hi, could you [sign off ](https://github.com/CycloneDX/cyclonedx-dotnet/pull/933/checks?check_run_id=38362264686) your commits, then I can merge and release your changes.

Hi, Yes, that is done now :)

PR #934, opened 2025-03-10
fix: issue 894 - remove dependencies, that have no belonging library

mtsfoni:
@Falco20019 This is my intended fix for issue #894. I simply kick out dependencies that have no corresponding library in the assets file as described as step 1 previously [here](https://github.com/CycloneDX/cyclonedx-dotnet/pull/929#issuecomment-2692962569).

This should work for you, shouldn't it?
A future step then would be to still find those dev-dependencies in a recursive scan of package-references. 

I'd appreciate a short review, otherwise I will just release it tomorrow over the day.


Falco20019:
I will have a look tomorrow and give you feedback. Already out of office for today.

Falco20019:
Only had time for a quick look but looks as it should work for me.

PR #939, opened 2025-04-07
Add exclude filter option

t-castelan:
We use CycloneDX-Dotnet for creating an SBOM for a project which targets .NET 8. Some transient dependencies are fairly old like [Antlr3.Runtime](https://www.nuget.org/packages/Antlr3.Runtime), so .NETStandard 1.6.1 will be pulled into the resulting SBOM including many old libraries, which will not be used by our application. They do not exist in the binary build output, so these dependencies should be removed from the resulting SBOM.

We could not find any existing feature request for dependency filtering in CycloneDX-Dotnet or CycloneDX-CLI. There is only a command line option for filtering test projects and development dependencies, but no option to provide a custom filter. So we decided to implement a custom filter option by ourselves. Please consider this feature as an improvement for the official repository.

### DIFF
--- a/CycloneDX/Runner.cs
+++ b/CycloneDX/Runner.cs
@@ -228,8 +228,8 @@ namespace CycloneDX
             {
                 if (!string.IsNullOrEmpty(options.DependencyExcludeFilter))
                 {
-                    ExcludePackages(packages, options.DependencyExcludeFilter);
-                    RemoveOrphanedPackages(packages);
+                    ExcludeFilterHelper.ExcludePackages(packages, options.DependencyExcludeFilter);
+                    ExcludeFilterHelper.RemoveOrphanedPackages(packages);
                 }
             }
             catch (ArgumentException e)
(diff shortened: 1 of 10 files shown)


CycloneDX/Runner.cs
```diff
@@ -550,5 +568,37 @@ internal static void AddMetadataTool(Bom bom)
...
+            {
+                return;
+            }
+
+            packages.Remove(packageToExclude);
+            foreach (var (name, version) in actualPackageToExclude.Dependencies)
+            {
+                var dependencyToExclude = new DotnetDependency { Name = name, Version = version };
+                ExcludePackage(packages, dependencyToExclude);
+            }
```
mtsfoni, line 600: This is not quite right. 
What you do is that you cut off all transitive dependencies of the excluded dependency - as if it was a tree, and you cut the branch. 
However, we have a dependency graph, which means that dependencies can be connected to the root component on different paths - transitive dependencies that are also used by other dependencies that are not excluded must not be  removed.

Here is an example:

```
dotnet new console -n CliWithDependencies
cd CliWithDependencies
dotnet add package Microsoft.Extensions.Logging
dotnet add package Microsoft.Extensions.Configuration
```

Both packages, `Logging` and `Configuration` depend on `Microsoft.Extensions.Primitives`. If I call cdx with `-ef Microsoft.Extensions.Logging` the library `Microsoft.Extensions.Primitives` is excluded even though it is also a dependency of Configuration, which was not excluded.

So before removing, you need to check if those things are referenced somewhere else.

mtsfoni (changes requested):
I made one comment that needs to be addressed; otherwise, everything looks good.

mtsfoni:
No problem, your approach to eliminate the orphans looks good. I just refactored the algorithm for it a bit and added outputs, so users will see in the logs that/why thing were removed.

Thanks for your PR

PR #944, opened 2025-05-04
feat: add evidence collect option

malytskyy:
The license evidence collection may be used to collect license information from shipped files, like LICENSE.txt. This is particularly useful for packages, which have no license id provided, but rather information is provided in a file. Also even when the license id or name is known, it still might be a good idea to have the license information from the time of BOM creation.

The default mode is `None` which means no license evidence will be collected. The other options are `All` which collects all license evidence, even when the license id is known. Lastly, `Unknown` Collect license text only for components which have unknown license. This avoids collecting all license texts for the case when license text can be obtained otherwise (like MIT) and therefore reduces the BOM size. In contrast to the "All" mode, this mode will put license text into license block directly instead of evidence part.

### DIFF
--- a/CycloneDX.Tests/NugetV3ServiceTests.cs
+++ b/CycloneDX.Tests/NugetV3ServiceTests.cs
@@ -48,7 +48,7 @@ namespace CycloneDX.Tests
             };
             var mockGithubService = new Mock<IGithubService>();
             var nugetService = new NugetV3Service(null, mockFileSystem, cachePaths, mockGithubService.Object,
-                new NullLogger(), false);
+                new NullLogger(), false, EvidenceLicenseTextCollectionMode.None);
 
             var nuspecFilename = nugetService.GetCachedNuspecFilename("TestPackage", "1.2.3");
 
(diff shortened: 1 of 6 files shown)


mtsfoni:
Hi,
sorry for the late review.
Do you happen to have a NuGet example package that has no license id provided, and provide such a file? I'd like to conduct a real world test.

malytskyy:
hi, no worries :) 

here are some examples (let me know if you need more)

Duende.IdentityServer
- https://www.nuget.org/packages/Duende.IdentityServer/7.3.2/License

Extreme.Numerics
- https://www.nuget.org/packages/Extreme.Numerics

Microsoft.CognitiveServices.Speech
- https://www.nuget.org/packages/Microsoft.CognitiveServices.Speech/1.47.0-beta.0.357883

Microsoft.Data.SqlClient.SNI.runtime
- https://www.nuget.org/packages/Microsoft.Data.SqlClient.SNI.runtime/6.0.2/License

Microsoft.Graph
- https://www.nuget.org/packages/Microsoft.Graph/5.94.0/License

a good example is `Microsoft.Graph` - actually it's a MIT licensed library, but someone apparently forgot to set the key in the package manifest. 

Regards

PR #958, opened 2025-06-24
feat: add spec version option (#949)

### LINKED ISSUE #949: Option to generate different CycloneDX Output Versions
Hi, 

I have the requirement to produce CycloneDX Version 1.4 sboms but I would like to use the latest
version of cyclonedx-dotnet.

Would it be possible to specify the FileVersion with a command line option?

ben-hamida:
This option allows generating a BOM file with any supported specification version. Fixes #949 

I opted to use `string` as generic parameter to `Option` as opposed to the `SpecificationVersion` enum since it was a lot simpler. With the enum I would have to specify a custom parser to support the x.y format as opposed to the vx_y format used in the enum names. Also, I found no way to remove the automatically added allowed values that would be of the wrong format (although I think this might be possible in later version of the `System.Commandline` package).

The calls to `Json.Serializer.Serialize` and `Xml.Serializer.Serialize` internally call `BomUtils.GetBomForSerialization` which automatically downgrades the BOM if needed. So I think this should just work.

I didn't feel like there is a need for tests since this is just passing along a parameter, but let me know if you disagree and I'll add some.


ben-hamida:
I just realized my changes only sets the specVersion property in the document. I was somehow under the impression that the data was sent to the core library for SBOM generation but I see now that the actual generation is done in `Runner.cs`. Now I understand the [comment in #949](https://github.com/CycloneDX/cyclonedx-dotnet/issues/949#issuecomment-2934612431) about downgrading better. Setting the PR to draft for now.

ben-hamida:
@mtsfoni Hi! Just wanted to give a friendly ping. No rush, but wanted to check if there's any chance this will be looked at soon.

Maxx53:
Looking forward to this feature, thanks!

CycloneDX/Program.cs
```diff
@@ -60,7 +60,11 @@ public static Task<int> Main(string[] args)
                 description: "Select the BOM output format: auto (default), xml, json, or unsafeJson (relaxed escaping).",
                 getDefaultValue: () => OutputFileFormat.Auto
             );
-
+            var specVersion = new Option<string>(
+                aliases: new[] { "--spec-version", "-spv" },
+                description: "Which version of CycloneDX spec to use.",
+                getDefaultValue: () => SpecificationVersionHelpers.VersionString(SpecificationVersionHelpers.CurrentVersion));
```
mtsfoni, line 66: Can we move the default logic to the `Runner.cs`-File for better testability?

mtsfoni (changes requested):
I'd appreciate one small test to just show that it actually works. (Handing over "1.5" leads to resulting SBOM having 1.5 as Version),

CycloneDX/Program.cs
ben-hamida: I like the automatic documentation in the help output when specifying it here. But I agree that's nice to be able to test that the correct default value is used. I added the default value in the help output manually as part of the description property and made  `RunOptions.specVersion` nullable while setting the default value in `Runner`. Added two tests, one for checking that the specified spec version is used and one that verifies that the correct default version is used.

mtsfoni (approved):

PR #985, opened 2025-11-20
NugetService: Find PackageVulnerabilityInfo from IVulnerabilityInfoResource (Issue 805)

LarsPellarin:
NugetService: Add support to find PackageVulnerabilityInfo from IVulnerabilityInfoResource.
Reduce code duplication in test and make it easier to inject vulnerability service mock.
Refactor to inject source repository from factory.

### DIFF
--- a/CycloneDX/Runner.cs
+++ b/CycloneDX/Runner.cs
@@ -317,7 +317,16 @@ namespace CycloneDX
                         {
                             vulnerabilities.AddRange(nugetComponent.Vulnerabilities.Select(v => new Vulnerability
                             {
-                                Source = new Source { Name = "NuGet", Url = v.AdvisoryUrl },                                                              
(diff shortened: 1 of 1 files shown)


mtsfoni:
@LarsPellarin  Cool that you took on that feature!

The Class from the CycloneDX-library shouldn't be used in the NugetService imo.
I replaced it with a dto class for the service and mapped it to the Library class in the runner.cs.

I think it still needs to use more fields of the Vulnerability and probably testing (log4net is a classic here, oracle adapter also has vulnerabilities). Let me know if you still want to on, otherwise I will finish it in the coming days.

LarsPellarin:
Thanks for feedback! Agree on the DTO refactor.

Feel free to finish this, i will not have time the next couple of days myself.

PR #999, opened 2025-12-16
Populate url when processing license expressions

giovanni-bozzano:
When a NuGet package defines its license by expression type and then provides the licenseUrl, the latter is missed during the collection.

This change applies the optional licenseUrl to all licenses processed from the expression, in order not to lose the information.


Falco20019:
@mtsfoni In #1076 you wrote that `expression + licenseUrl` is invalid. That's correct, but that's also not what's happening in this PR. This PR uses the license metadata to get the URL, not the nupkg itself. For nuget.org this will contain the URL as `https://licenses.nuget.org/<expression>`.

We now went on to use the `Spdx3` package to resolve the ID to license texts instead of hoping that it's set by the SBOM. But for customers this would still be a benefit to know where the license text can be fetched from.

PR #1044, opened 2026-03-01
feat!: always emit dev dependencies with scope=excluded; deprecate --exclude-dev

mtsfoni:
## Summary

Dev dependencies are now always included in the BOM with `scope="excluded"` rather
than being silently dropped when `--exclude-dev` is passed. This aligns with the
CycloneDX spec, which expects excluded components to be present but marked as such.

## Changes

- **`--exclude-dev` deprecated** — the flag is retained for backward compatibility
  but is now a no-op. Passing it emits a deprecation warning. Planned for removal
  in a future major release.
- **BFS-based scope resolution** — replaces the old eager flag-and-filter approach.
  After all packages are loaded, a BFS from direct runtime references determines
  reachability. Packages never reached via a runtime path receive `scope=excluded`.
  This correctly handles the shared-transitive case: a package reachable from both
  a dev dep and a runtime dep is kept in scope.
- **`ProjectAssetsFileService`** — dev dependency scope is no longer set eagerly;
  test-project exclusions are unaffected.

## Breaking Change

Previously, `--exclude-dev` omitted dev dependency entries from the BOM entirely.
They are now always present with `scope="excluded"`. Consumers that relied on the
flag to suppress those entries from output will see additional components.

## Testing

- Two new functional tests: transitive-only-via-dev and shared-transitive scenarios.
- Existing dev dependency tests updated to reflect the new always-included behavior.
- New `docs/dev-and-test-dependencies.md` documents the behavior and rationale.


gotztibor:
Dear Michael, would You complete this PR? This would be very useful for us to correctly track the dev dependencies.

